Issue #7, opened 2024-06-17
APEx benchmarks: add metrics


JeroenVerstraelen:
Add metrics to the initial benchmark tests.

- For each benchmark, extract metrics using openEO 
  - Preferably relative metrics: 
     - cost per produced sqkm
     - cost per input pixel
  - Timing related metrics
  - Discuss which other metrics we can include 
- Upload results as geoparquet file to S3 object storage
  - Which endpoint/bucket? 

Goal: partitioned parquet file (partitioned on: time based, benchmarking scenarios)

soxofaan:
added some docs at https://github.com/ESA-APEx/apex_algorithms/blob/main/docs/benchmarking.md

soxofaan:
most important aspect to still cover under this ticket: 
- [ ] record collected metrics in a parquet file on S3

soxofaan:
First results from pushing some metrics as parquet to S3:

![image](https://github.com/user-attachments/assets/117e688b-4928-41dc-b016-4c99a3016c72)

soxofaan:
and now with unrolling the usage stats:
![image](https://github.com/user-attachments/assets/4e5d58eb-5a7a-415d-ad9f-8ce3f87dcc72)

soxofaan:
merged PR #26 which covers the part about partitioned parquet files on S3

soxofaan:
Ok I think it's time to close this ticket.
There were quite some aspects to it and I'm still unsure about some parts of the current approach, so this feels mostly like a proof of concept solution. Also, not everything of this ticket's original requirements are met, but those were "TBD" anyway.

Some details and discussion about this PoC:
- benchmarks are implemented as pytest test suite
- metrics are collected through a pytest fixture `track_metric` provided by pytest plugin `apex_algorithm_qa_tools.pytest.pytest_track_metrics`. Currently collected metrics/properties:
    - scenario_id, job_id
    - job costs
    - job's usage metrics (e.g. usage:cpu:cpu-seconds, usage:memory:mb-seconds)
    - test info: outcome (passed/failed), start time, duration
- these metrics are written in parquet format to S3 (bucket "APEx-benchmarks"), in a folder structure starting at "metrics/v1/metrics.parquet"
    - current partitioning is based on month of the benchmark run, e.g. folder "metrics/v1/metrics.parquet/2024-08/"
[...]

Issue #5, opened 2024-06-17
APEx benchmarks: add regression test


JeroenVerstraelen:
Add regression tests to the initial benchmarks.

- To generate reference data, run benchmark once
- Upload reference data to s3 object storage
  - What endpoint/bucket?
- From then on every benchmark run should also compare its output to the reference data
- Alerting when output does not match?
  - For the UDP owner, openEO team? 
  - How to configure?
- Save comparison result (Success, failure) to s3 object storage
  - So we can see a timeseries for each benchmark 
  - Which endpoint/bucket?

Requirements:
- Should be easy to update reference data.
- Component that we can reuse in our own integrationtests. 
- Mostly via configuration, try to avoid having to write custom code for every scenario.

soxofaan:
> Component that we can reuse in our own integrationtests.

I started with reusable job result comparison utilities in the python client at https://github.com/Open-EO/openeo-python-client/pull/587

soxofaan:
merged https://github.com/Open-EO/openeo-python-client/pull/587 no

JeroenVerstraelen:
Sync with Thomas for s3 buckets.

soxofaan:
got s3 bucket info, and set up initial poc for comparing job results with reference data: bf11993827fd7f9d990724dbc4eaf0fc787d6207

soxofaan:
First (intendedly) failing benchmark using reference data on S3 and brand new (unreleased) `assert_job_results_allclose` functionality of python client:
https://github.com/ESA-APEx/apex_algorithms/actions/runs/9961442215/job/27522945120

```
E           AssertionError: Issues for file 'openEO.tif':
E           Coordinates mismatch for dimension 'x': [295355. 295365. 295375. 295385. 295395. 295405. 295415. 295425. 295435.
E            295445. 295455. 295465. 295475. 295485. 295495. 295505. 295515. 295525.
...
E            296705. 296715. 296725. 296735. 296745. 296755. 296765. 296775. 296785.
E            296795. 296805. 296815. 296825. 296835.] != [644525. 644535. 644545. 644555. 644565. 644575. 644585. 644595. 644605.
E            644615. 644625. 644635. 644645. 644655. 644665. 644675. 644685. 644695.
...
E            645785. 645795. 645805. 645815. 645825. 645835. 645845. 645855. 645865.
E            645875. 645885. 645895. 645905. 645915. 645925. 645935. 645945. 645955.
E            645965. 645975. 645985.]
[...]

PR #19, opened 2024-07-22
HTML+JSON reporting

soxofaan:
Part of
- #5 
-  #7


qa/tools/apex_algorithm_qa_tools/test_metrics.py
```diff
@@ -0,0 +1,126 @@
+"""
```
soxofaan, line 1: `test_` is kind of a reserved prefix in pytest, 
pick something better, with less risk of confusion/collision?

qa/tools/apex_algorithm_qa_tools/test_metrics.py
```diff
@@ -0,0 +1,126 @@
...
+Pytest plugin to record test/benchmark metrics to a JSON file.
```
soxofaan, line 2: also add bit of testing for this plugin

qa/tools/apex_algorithm_qa_tools/test_metrics.py
soxofaan: renamed to "track_metrics"

qa/tools/apex_algorithm_qa_tools/test_metrics.py
soxofaan: done

.github/workflows/benchmarks.yaml
```diff
@@ -21,7 +21,14 @@ jobs:
       - name: Test with pytest
         run: |
           cd qa/benchmarks
-          pytest --random-subset=1
+          mkdir report
+          pytest -k dummy --html report/report.html --self-contained-html --test-metrics=report/metrics.json
```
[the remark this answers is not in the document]
soxofaan: done

Issue #15, opened 2024-07-01
Merge esa-apex-toolbox-python repo


soxofaan:
Under https://github.com/ESA-APEx/apex_algorithms/issues/3 I created a new repo https://github.com/ESA-APEx/esa-apex-toolbox-python

But while working on https://github.com/ESA-APEx/apex_algorithms/issues/4 I started to realize that there is a lot of common ground between [ESA-APEx/apex_algorithms](https://github.com/ESA-APEx/apex_algorithms)  and  [ESA-APEx/esa-apex-toolbox-python](https://github.com/ESA-APEx/esa-apex-toolbox-python), e.g. the schemas of resources, how to parse/validate them, ... 

At the moment there seems to be emerging some kind of bi-directional dependency between these projects and I think it makes more sense (development/maintenance-wise) to just merge esa-apex-toolbox-python as a subproject into apex_algorithms. Especially at this early stage of development.

soxofaan:
As discussed over at https://github.com/ESA-APEx/apex_documentation/issues/69  we're now having broken builds because of hidden and now broken coupling between these projects

PR #22, opened 2024-07-26
Hv addprofilemapping

HansVRP:


jdries:
@HansVRP  I ran the notebook myself, on my max ndvi example:

1.  I made small improvements based on that experience
2. The heatmap has an issue, spatial and temporal range seem to be expected in the dataframe, but weren't there in my case
3. I would avoid adding NDWI udp
4. We'll want to integrate the code perhaps with python library

Otherwise, this is a great tool, and would be good to get it merged so we can include it in documentation and SRR.

HansVRP:
@jdries Hi Jeroen this was really a dummy example. Let's discuss the requisites for the underlying UDPs and then I can properly expand the benchmark/visualisation to cover all requirements

jdries:
Hi @HansVRP , the goal for SRR in octobre is to have an MVP that is sufficiently visible to the reviewers. So in terms of functionality, this is already enough, we just need to increase visibility by putting it in apex docs and out of a pr.
It is fine to clearly label it as MVP/prototype, to avoid bad experiences.

If fixing the heatmap seems to complex, we can also just not use it, or show a screenshot.

HansVRP:
@jdries should be good for review. Stepped away from the class based implementation and cleaned up the notebook to make it more explanatory

jdries:
ok:
ndwi.json is probably not necessary anymore?

There's this issue:
https://github.com/ESA-APEx/apex_algorithms/issues/15

So not sure what to do, just merge this one and then in follow up work merge both projects and also move this code to a better location?

HansVRP:
that sounds indeed like a good follow-up step

PR #27, opened 2024-09-06
Max ndvi composite

jdries:
new udp to generate max ndvi composite, to serve as a full example for other udp's


openeo_udp/examples/max_ndvi_composite/generate.py
```diff
@@ -0,0 +1,105 @@
...
+        "unclassified": 7,
+        "cloud_medium_probability": 8,
+        "cloud_high_probability": 9,
+        "thin_cirrus": 10,
+        "snow": 11,
+    }
+
+    from openeo.processes import any
+
+    def scl_to_masks(bands):
```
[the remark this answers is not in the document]
jdries: I'll try to do that. It has some potential in improving the performance, and indeed avoids that users need to apply a water mask themselves.

HansVRP (approved):

openeo_udp/examples/max_ndvi_composite/generate.py
```diff
@@ -0,0 +1,118 @@
...
+    scl = c.load_collection(
+        "SENTINEL2_L2A",
+        temporal_extent=temporal_extent,
+        bands=["SCL"],
+        max_cloud_cover=max_cloud_cover_param
+    ).resample_spatial(projection="EPSG:25832", resolution=10)
+
+    def scl_to_masks(bands):
+        scl_layer = bands[0]
+        to_mask = any(array_create([
```
soxofaan, line 42: I would use `openeo.processes.any` here instead of just `any`, it's very easy to miss the `from openeo.processes import any` lower and think it's python's standard `any` at work here

openeo_udp/examples/max_ndvi_composite/generate.py
```diff
@@ -0,0 +1,118 @@
...
+
+    scl = c.load_collection(
+        "SENTINEL2_L2A",
+        temporal_extent=temporal_extent,
+        bands=["SCL"],
+        max_cloud_cover=max_cloud_cover_param
+    ).resample_spatial(projection="EPSG:25832", resolution=10)
+
+    def scl_to_masks(bands):
+        scl_layer = bands[0]
```
soxofaan, line 41: There is something wrong here: `scl_to_mask` is used as `apply` callback, so the argument to `scl_to_mask` has to be considered to be a single scalar, while the `bands[0]` (syntactic sugar for array_element) here indicates it is assumed to be an array

algorithm_catalog/max_ndvi_composite_description.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Performance characteristics
+
+The method is computationally efficient, as it only requires the B04, B08 and SCL bands to determine the rank score. Loading 
+of other bands can be minimized to read only selected observations.
+
+
+# Examples
+
+The image below shows a typical result over an agricultural area.
```
Patrick1G, line 19: I suggest to add: "Note the typical artifacts due to issues with the S2 SCL cloud masking."

algorithm_catalog/max_ndvi_composite_description.md
```diff
@@ -0,0 +1,71 @@
...
+The method falls under the 'rank composite' category, and ensures that selected spectral band values for any individual pixel all come from the same observation.
+
+The method generates good results for timeseries analytics, but spatially neighbouring pixels may be selected from different observations,
+which leads to visual discontinuities in the result.
+
+# Performance characteristics
+
+The method is computationally efficient, as it only requires the B04, B08 and SCL bands to determine the rank score. Loading 
+of other bands can be minimized to read only selected observations.
+
```
Patrick1G, line 15: Can we say anything regarding maximum AOI and/or timeframe? This is a key performance consideration for openEO UDPs...

algorithm_catalog/max_ndvi_composite_description.md
```diff
@@ -0,0 +1,71 @@
...
+Credits: 63 
+CPU usage: 47.743,722 cpu-seconds
+Wall time: 1.948 seconds
+Input Pixel 10.997,635 mega-pixel
+Max Executor Memory: 3,239 gb
+Memory usage: 154.767.121,977 mb-seconds
+Network Received: 1.677.537.930.040 b
+```
+
+The relative cost is 1 platform credits per km² for a 3 month input window.
```
Patrick1G, line 39: is this an openEO platform credit (also CDSE?)? might be worth noting, as other credits might not be comparable...

algorithm_catalog/max_ndvi_composite_description.md
```diff
@@ -0,0 +1,71 @@
...
+The method uses a vegetation index as scoring metric to determine the best pixel, making it only suitable for land applications. 
+Bare or urban areas may not be well represented in the composite. 
+
+It favours the observation which is least contaminated by atmospheric effects, but does not guarantee a fully uncontaminated composite.
+
+For individual time windows, the method was efficient up to 100x100km areas. For larger areas of interest, we recommend splitting the area into smaller tiles.
+
+
+# Known artifacts
+
```
Patrick1G, line 63: those are a few heavy artifact examples, maybe discouraging if we do not show a beautiful and convincing examples as well ;-)

Patrick1G: another thought: current the UDP does not produce any ancillary layers like DOY of chosen observation or number of cloud free observations. These can be quite valuable. Does your algo produce these?

openeo_udp/examples/max_ndvi_composite/generate.py
jdries: actually it probably works both, but I adjusted it

algorithm_catalog/max_ndvi_composite_description.md
jdries: indeed it follows a bit later under 'known limitations'

algorithm_catalog/max_ndvi_composite_description.md
jdries: I added cdse explicitly. The link with the platform is also made in the metadata json.
Ideally, when the web portal renders our algorithm catalog, it should clearly show which algorithms are (officially) associated with which platform.
Also good to note that this algorithm will in fact work just fine on openEO platform, only thing is that SENTINEL2_L2A might come in via sentinelhub from aws, which fill inflate credit cost a lot.

algorithm_catalog/max_ndvi_composite_description.md
jdries: The example I added under 'examples' and also the large geotiff I reference do look a bit nicer.
As I was creating this as an example for other UDP writers, I was also trying to be very honest about such issues, hoping to stimulate others to also be somewhat transparent on known issues.

Current method indeed does not store date of selected observation. Personally, if I need to know the selected observations, I just drop the max_time (reduce_temporal) at the end and write a multitemporal netcdf. 
Anyway, for now I suggest to keep it simple, and see how we integrate further requests?

jdries:
I created a composite over 50x100km, accidentally over 15 months, 3 output bands.
Credit cost: 189 (that's not too much)

Main goal was to try with masking over water. It reduces some of the ugliness, but unfortunately there's also water pixels that are classified as something else, thus remaining in the final composite:

image

jdries:
I'm going to merge this one, allowing me to move on.

PR #39, opened 2024-10-09
GEP Services Examples

p3dr0:
Here are two examples of services instantiated in GEP


jdries:
Hi @p3dr0, thanks, these records indeed look valid.
I propose that we do a more thorough review of these first cases, so that we have at least 1 high quality example that we can point other users to. (That's how we did it for openEO as well.)

So as for review:

The description/documentation is important, here's an example of a description we did:
https://github.com/ESA-APEx/apex_algorithms/blob/main/algorithm_catalog/max_ndvi_composite_files/max_ndvi_composite_description.md
I think we now put that in the udp, but it can also be used at the level of the OGC record.

Of course, if you have good documentation on the GEP, it makes a lot of sense to also link to that.

What I also did, is adding a link to STAC metadata of example output, like this:
```
{
  "rel": "example",
  "type": "application/json",
  "title": "Example output",
  "href": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-examples/max_ndvi_denmark/collection.json"
}
```
I think this is one of the most convenient ways to show what an algorithm actually produces, and this is something that we can then for instance show in geospatial explorer.

Finally, but most importantly, can you add the link to the OGC Process description? This is the crucial element for all the tooling to work, but also simply required for users to be able to generate output.

JanssenBrm:
@p3dr0  - Have you already had the chance to look at the previous request for updates? It may be too late to fully integrate the OGC Processes in all of the tools, but it would be nice if we could (at least) show these services in our catalogue.  Our catalogue team is waiting for the finalization of this PR to continue the work.

p3dr0:
I updated both json files with links
 
please note that I'm considering these relations types 

- “rel”: “order” -> is where you can obtain information about how to buy it
- “rel”: “license”, -> link for the EULA of the service
- “rel”: “service”, -> direct link to submit, status, etc … (no need to make string addition with other rels)
- “rel”: “documentation” -> direct link for the documentation pages
- “rel”: “application” -> direct links for the application package

these are self-standing, this means from each alone you are able to complete an action in a given context (provided by the relation and mime-type), it should not be needed to sum them together in pairs to provide an action  

in the examples updated, both the service and application links are protected so you'll get a 403 for now

PR #44, opened 2024-10-16
Hv pv detection

HansVRP:
added EURAC UDP. 

There are some points which are still open for debate: 

- Do we want to set the max cloud cover as a parameter as well and/or the kernel size of the postprocessing filter?

- Currently I get the model and dependencies within the UDF, I saw that for world cereal job options are given. Is there a preferred approach?


soxofaan:
> Currently I get the model and dependencies within the UDF, I saw that for world cereal job options are given. Is there a preferred approach?

at the moment there is no standard way to attach job options to UDPs, e.g see this discussion:
- https://github.com/Open-EO/openeo-api/issues/545

so for the moment, I think doing it in the UDF is the only way

HansVRP:
In the word cereal benchmark they are passed as: 

 "job_options": {
        "driver-memory": "4g",
        "executor-memory": "1500m",
        "python-memory": "5g",
        "udf-dependency-archives": ["https://artifactory.vgt.vito.be/artifactory/auxdata-public/openeo/onnx_dependencies_1.16.3.zip#onnx_deps"]
    }
  }
  
  hence my confusion.
  
  Is this then different?

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,156 @@
...
+# Ensure directories exist
+os.makedirs(dependencies_dir, exist_ok=True)
+os.makedirs(model_dir, exist_ok=True)
+
+# Download and extract dependencies and model if not already present
+if not os.listdir(dependencies_dir):
+    download_and_extract(dependencies_url, dependencies_dir)
+    
+if not os.listdir(model_dir):
+    download_and_extract(model_url, model_dir)
```
soxofaan, line 42: Isn't it possible to encapsulate all this toplevel makedirs/listdir/download_and_extract stuff in a function? If so, then it would make sense to actually move that kind of functionality to the python client in some kind of UDF utility module.

soxofaan (commented):
some notes

soxofaan:
> In the word cereal benchmark they are passed as: ...

ah ok I see what you mean now.

you can indeed define job options in the benchmark scenario

there are indeed two options at the moment
- handle model/deps "manually" from the UDF code. Adds more boilerplate to the UDF, but makes usage  easier  (less job_option worries)
- work with job options. less UDF boilerplate, but as a user you're forced to always set job options. And there is unfortunately no standard way jet to attach "recommended" job options to a UDP

I'm not sure what is best at the moment

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,156 @@
...
+
+# Add dependencies to the Python path
+sys.path.insert(0, dependencies_dir)
```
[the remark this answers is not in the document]
HansVRP: Probably not, I think it is some leftover code from when I needed to add a different version of package  than standardly available

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
HansVRP: I transfer the code to MLconf.py, should we make a seperate issue to tackle this part independently?

HansVRP: I will make the code more modular, but maybe its best if we create a seperate issue for this?

openeo_udp/eurac_pv_farm_detection/generate.py
```diff
@@ -0,0 +1,111 @@
...
+
+    #load the input data
+    conn = openeo.connect("https://openeofed.dataspace.copernicus.eu/").authenticate_oidc()
+
+    s2_cube = conn.load_collection(
+            collection_id="SENTINEL2_L2A",
+            spatial_extent=spatial_extent,
+            temporal_extent=temporal_extent,
+            bands=["B02", "B03", "B04", "B05", "B06", "B07", "B08", "B8A", "B11", "B12"],
+            max_cloud_cover=85)
```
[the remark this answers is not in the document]
HansVRP: Ran Isort and black

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def download_file(url, path):
+    """
+    Downloads a file from the given URL to the specified path.
+    """
+    response = requests.get(url, stream=True)
+    with open(path, "wb") as file:
+        file.write(response.content)
+    print(f"Downloaded file from {url} to {path}")
```
soxofaan, line 25: UDFs should not use print. I don't think the user is able to see that anyway. And I even think this has potential to break the batch job logging.

In any case, we should not promote using "print" from UDFs

benchmark_scenarios/eurac_pv_farm_detection.json
```diff
@@ -0,0 +1,27 @@
+[
+    {
+      "id": "eurac_pv_farm_detection",
+      "type": "openeo",
+      "description": "ML photovoltaic farm detection, developed by EURAC",
+      "backend": "openeofed.dataspace.copernicus.eu",
+      "process_graph": {
+        "maxndvi1": {
+          "process_id": "eurac_pv_farm_detection",
+          "namespace": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/hv_pv_detection/openeo_udp/eurac_pv_farm_detection/eurac_pv_farm_detection.json",
```
soxofaan, line 10: Note that this reference will break once the `hv_pv_detection` feature branch is removed

also see https://github.com/ESA-APEx/apex_algorithms/issues/48

algorithm_catalog/eurac_pv_farm_detection.json
```diff
@@ -0,0 +1,104 @@
...
+    ],
+    "license": "CC-BY-4.0"
+  },
+  "linkTemplates": [],
+  "links": [
+    {
+      "rel": "openeo-process",
+      "type": "application/json",
+      "title": "openEO Process Definition",
+      "href": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/hv_pv_detection/openeo_udp/eurac_pv_farm_detection/eurac_pv_farm_detection.json"
```
soxofaan, line 85: Note that this reference will break once the `hv_pv_detection` feature branch is removed

also see https://github.com/ESA-APEx/apex_algorithms/issues/48

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,196 @@
...
+        download_file(model_url, zip_path)
+        extract_zip(zip_path, MODEL_DIR)
+        add_directory_to_sys_path(MODEL_DIR)
+
+
+DEPENDENCIES_URL = "https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies/onnx_dependencies_1.16.3.zip"
+MODEL_URL = "https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies/EURAC_pvfarm_rf_1_median_depth_15.zip"
+
+
+setup_model_and_dependencies(model_url=MODEL_URL, dependencies_url=DEPENDENCIES_URL)
```
soxofaan, line 77: I don't think it's necessary to define  DEPENDENCIES_URL and MODEL_URL as top-level constants.

This should also work I think:
```
setup_model_and_dependencies(
    model_url="https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies/EURAC_pvfarm_rf_1_median_depth_15.zip",
    dependencies_url="https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies/onnx_dependencies_1.16.3.zip",
)
```

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,196 @@
...
+    Make sure that the arguments of the method you add the @functools.lru_cache decorator to are hashable.
+    Be careful with using this decorator for class methods, as the self argument is not hashable.
+    In that case you can use a static method or make sure your class is hashable (more difficult): https://docs.python.org/3/faq/programming.html#faq-cache-method-calls.
+
+    More information on this functool can be found here:
+    https://docs.python.org/3/library/functools.html#functools.lru_cache
+    """
+    # The onnx_models folder contains the content of the model archive provided in the job options
+    return ort.InferenceSession(
+        f"model_files/{model_name}", providers=["CPUExecutionProvider"]
```
soxofaan, line 102: This path should refer to the `MODEL_DIR` constant

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # Add the extracted dependencies directory to sys.path
+        add_directory_to_sys_path(DEPENDENCIES_DIR)
+
+    # Download and extract model if not already present
+    if not os.listdir(MODEL_DIR):
+        zip_path = os.path.join(MODEL_DIR, "temp.zip")
+        download_file(model_url, zip_path)
+        extract_zip(zip_path, MODEL_DIR)
+        add_directory_to_sys_path(MODEL_DIR)
```
soxofaan, line 70: Why is the model dir added to sys path? That shouldn't be necessary, right?

benchmark_scenarios/eurac_pv_farm_detection.json
HansVRP: Yes, so is the idea that once its merged we do a follow-up PR? Or is there an alternative solution?

soxofaan: I think, if you use a "permalink", it will keep working after the feature branch is removed

in this case that would be 
> https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/4003046e3b79ec3ab8dace888a231655db389d66/openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py

it's a bit cumbersome to explain how to get that from within this comment, but we can discuss this elsewhere

algorithm_catalog/eurac_pv_farm_detection.json
```diff
@@ -82,7 +82,7 @@
       "rel": "openeo-process",
       "type": "application/json",
       "title": "openEO Process Definition",
-      "href": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/hv_pv_detection/openeo_udp/eurac_pv_farm_detection/eurac_pv_farm_detection.json"
+      "href": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/4003046e3b79ec3ab8dace888a231655db389d66/openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py"
```
soxofaan, line 85: sorry for being vague earlier, I just gave an example permalink, I think you still want the JSON here (you changed it to `.py`)

benchmark_scenarios/eurac_pv_farm_detection.json
```diff
@@ -7,7 +7,7 @@
       "process_graph": {
         "maxndvi1": {
           "process_id": "eurac_pv_farm_detection",
-          "namespace": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/hv_pv_detection/openeo_udp/eurac_pv_farm_detection/eurac_pv_farm_detection.json",
+          "namespace": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/4003046e3b79ec3ab8dace888a231655db389d66/openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py",
```
soxofaan, line 10: same here

algorithm_catalog/eurac_pv_farm_detection.json
soxofaan: I think you only have to replace the `refs/heads/hv_pv_detection` part from the URL to `4003046e3b79ec3ab8dace888a231655db389d66` (which is the commit you want to point to)

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,189 @@
...
+import zipfile
+from typing import Dict
+
+import numpy as np
+import requests
+import xarray as xr
+
+
+def _setup_logging():
+    logging.basicConfig(level=logging.INFO)
```
soxofaan, line 14: I don't think we should promote calling `logging.basicConfig` from UDF code. Backend-side it won't do anything, so this sets wrong expectations

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,189 @@
...
+    logger = logging.getLogger(__name__)
+    return logger
```
soxofaan, line 16: Instead of using `logger`, I would promote usage of `inspect` (https://open-eo.github.io/openeo-python-client/udf.html#udf-logging-with-inspect)

openeo_udp/eurac_pv_farm_detection/udf_eurac_pvfarm_onnx.py
```diff
@@ -0,0 +1,189 @@
...
+    # Ensure base directories exist
+    os.makedirs(DEPENDENCIES_DIR, exist_ok=True)
+    os.makedirs(MODEL_DIR, exist_ok=True)
+
+    # Download and extract dependencies if not already present
+    if not os.listdir(DEPENDENCIES_DIR):
+        logger.info("Extract dependencies")
+        zip_path = os.path.join(DEPENDENCIES_DIR, "temp.zip")
+        download_file(dependencies_url, zip_path)
+        extract_zip(zip_path, DEPENDENCIES_DIR)
```
soxofaan, line 69: This is probably out of scope of this ticket, but I think there is a race condition problem here: because of non-zero time between the `os.listdir(DEPENDENCIES_DIR)` check and actually writing the download to file, there could be multiple executors doing the same download or unzip, potentially causing corruption

HansVRP: Indeed.

would it help to cache the extraction functionality?

soxofaan: no, caching (I assume you mean in-process caching) won't help because the problem lies in concurrency between different executors (different processes possibly on different hosts).

A solution would be to work with a lock file to ensure ownership before starting the download. But as mentioned, this is probably out of scope of this ticket, and probably better handled when incorporating this helper functionality in the python client

Issue #45, opened 2024-10-17
Start enforcing pre-commit?


soxofaan:
I added a pre-commit config pretty early in June to get some basic static code quality management in place. At the moment using pre-commit is optional (opt-in) for developers and I noticed that quite some code style issues are slipping in the repo

e.g. after running `pre-commit run --all-files` to fix various issue, I get these diff stats currently:

```
$ git diff --stat 
 algorithm_catalog/max_ndvi_composite.json          |  11 +-
 .../max_ndvi_composite_description.md              |  10 +-
 benchmark_scenarios/max_ndvi_composite.json        |   4 +-
 benchmark_scenarios/worldcereal.json               |  15 +-
 openeo_udp/examples/max_ndvi_composite/generate.py |  84 +++++---
 .../max_ndvi_composite/max_ndvi_composite.json     |   2 +-
 openeo_udp/worldcereal_inference.json              |   6 +-
 openeo_udp_cost_profiling/utils.py                 | 220 ++++++++++++---------
 qa/unittests/tests/test_algorithm_catalog.py       |   5 +-
 9 files changed, 212 insertions(+), 145 deletions(-)
``` 

While I understand that playing pedantic code style police is not the point of this project, I still think this project plays an exemplary role in the ecosystem and we should set some bar related to code quality and give nice and consistent examples.

I think we can start with running a pre-commit check as github action to flag new issues

soxofaan:
I already addressed the reported issues on main branch and created PR #46 to add it as github action on PRs

PR #46, opened 2024-10-17
Introduce  pre-commit github action workflow

soxofaan:
For issue #45: this PR introduces a pre-commit github action workflow (e.g. to run on PRs)

todo
- [ ] don't run all hooks, just the most important ones, to not be too pedantic


soxofaan:
>   don't run all hooks, just the most important ones, to not be too pedantic

:disappointed:   I'd hoped there would be an option to run a subset, but it's only possible to run all hooks (default) or a single specific one

PR #37, opened 2024-10-03
Biopar vito porting to APEX

Pratichhya:
The UDP build, published within the Copernicus Data Space Ecosystem, has been modified to include only the JSON file, with the generate.py script excluded.

Additionally, as per the document's requirements, the repository link is not publicly available and therefore has not been included.

PR #68, opened 2024-12-02
Biopar masklayer

Pratichhya:
Updated the masking function in the Biopar process graph.

However, there still seems to be unittest failure, particularly in the following files:

- apex_algorithms/algorithm_catalog/combi_expressions/sar_water_mask.json
- apex_algorithms/algorithm_catalog/sar_coin.json
- apex_algorithms/algorithm_catalog/advanced_band_combination.json

@jdries, do you suggest waiting for the above files before merging this PR? So that I can pull from the main again. 

Please kindly let me know if I missed any points to correct 😅


jdries:
For spatial filtering, is there a reason not to follow this guideline?
https://esa-apex.github.io/apex_documentation/guides/udp_writer_guide.html#spatial-filtering

jdries:
That suggestion was indeed confusing, but let's use the guidelines as reference. If there are reasons to change it, we should update that first.

openeo_udp/biopar/biopar.json
```diff
@@ -59,8 +59,8 @@
             }
         },
         {
-            "description": "This geometries defines the bounds of the resulting image. Pixels outside the polygon are set to nodata.",
-            "name": "geometries",
+            "description": "This spatial_extent defines the bounds of the resulting image. Pixels outside the polygon are set to nodata.",
```
soxofaan, line 62: This is a bit confusing:  "defines the bounds" seems to suggest that `spatial_extent` is a bounding box construct (like in the prototypical `load_collection` usage), but the type  of this parameter is "geojson", so a bbox isn't even valid

openeo_udp/biopar/biopar.json
```diff
@@ -73,83 +73,119 @@
...
+        "filterspatial1": {
+            "process_id": "filter_spatial",
             "arguments": {
                 "data": {
-                    "from_node": "loadcollection1"
+                    "from_node": "filtertemporal1"
                 },
-                "extent": {
-                    "from_parameter": "temporal_extent"
+                "spatial_extent": {
```
soxofaan, line 121: the argument of `filter_spatial` is "geometries", not "spatial_extent"
```suggestion
                "geometries": {
```

openeo_udp/biopar/biopar.json
```diff
@@ -73,83 +73,119 @@
...
+                }
+            }
+        },
+        "filterspatial2": {
+            "process_id": "filter_spatial",
+            "arguments": {
+                "data": {
+                    "from_node": "filtertemporal2"
+                },
+                "spatial_extent": {
```
soxofaan, line 154:
```suggestion
                "geometries": {
```

soxofaan:
> thought to stick to this suggestion: https://github.com/ESA-APEx/apex_algorithms/pull/37#discussion_r1799341965 , no other specific reason

FYI: in that comment I just wanted to mention that "polygon" is not an argument name we use in standard processes, but "geometries" is (e.g. in `filter_spatial`, `aggregate_spatial`)


About the guidelines on spatial filtering: I think there is room for clarification: 

> In other openEO processes, the spatial filtering is called `spatial_extent`,

`spatial_extent` is only used in `load_collection` and `load_stac`, and there it accepts: bbox construct, inline geojson and vector cubes

In the use case here (openeo_udp/biopar/biopar.json), the parameter `spatial_extent` is defined as geojson-only (no bbox-support) and is passed to `filter_spatial` process (which also doesn't support bbox filtering). 

Wouldn't it be better to define the `spatial_extent` parameter as bbox+geojson+vectorcube and pass it to `load_collection` (instead of `filter_spatial`?

jdries:
> Wouldn't it be better to define the `spatial_extent` parameter as bbox+geojson+vectorcube and pass it to `load_collection` (instead of `filter_spatial`? The guidelines also recommend this

Yes, @Pratichhya  the idea is that you don't have to use filter_spatial anymore, parameter goes straight into load_collection.

I also suggest to use this convenience method to declare the parameter in Python:
https://open-eo.github.io/openeo-python-client/api.html#openeo.api.process.Parameter.spatial_extent

Pratichhya:
ahh, ok, will do that; directly load the extents in load_collection. Indeed, it is an easier solution. However, for Biopar in CDSE, it was suggested that the data be clipped to the spatial extent rather than bbox (remain the same as the VITO version), so I assumed the same might be expected here.

jdries:
@Pratichhya that's exactly what this parameter allows: clipping to bbox, geometry, vector cube, or even nothing at all and leaving it open to further downstream processing.

jdries:
Here's a PR for that spatial_extent confusion:
https://github.com/ESA-APEx/apex_documentation/pull/66

Pratichhya:
Thank you, @jdries and @soxofaan ; I just made updates as suggested above.

soxofaan:
about the failing tests: I already fixed some things on master branch, so I would merge master in this feature branch

soxofaan:
yes that wouldn't hurt to verify where test failures originate from (e.g. they might already be fixed in master)

Pratichhya:
The unittest seems to fail for:

- algorithm_catalog/sar_water_mask.json
- algorithm_catalog/snap_insar_sentinel1_iw_slc.json
- algorithm_catalog/sar_coin.json
- algorithm_catalog/gep_bas.json

also it shows some partitioning error not sure exactly.

Pratichhya:
> I'm not seeing those, what do you mean with that?

Sorry, I realised late that they were caused in my local settings and not related to the repo itself. When running the pytest locally.

Pratichhya:
@soxofaan, can I go ahead and merge this or it is suggested to change the test setting first

soxofaan:
yes I think it's fine to merge this

fixing the failing test on main branch is topic of 
- https://github.com/ESA-APEx/apex_algorithms/issues/76

so I guess we should assign that to somebody to work on

PR #72, opened 2024-12-06
clean up

HansVRP:
the included job options should resolve the failing integration test

### DIFF
--- a/benchmark_scenarios/eurac_pv_farm_detection.json
+++ b/benchmark_scenarios/eurac_pv_farm_detection.json
@@ -1,27 +1,36 @@
 [
-    {
-      "id": "eurac_pv_farm_detection",
-      "type": "openeo",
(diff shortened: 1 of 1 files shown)


HansVRP:
@soxofaan I notice that other unit tests are failing, I do not believe any are a result of my proposed change. Can we close this PR?

benchmark_scenarios/eurac_pv_farm_detection.json
```diff
@@ -1,27 +1,36 @@
...
       }
+    },
+    "job_options": {
+      "driver-memory": "1g",
+      "executor-memory": "1g",
+      "python-memory": "3g"}
+    ,
+    "reference_data": {
+    },
+    "reference_options": {
```
soxofaan, line 33: If you leave "reference_options" empty, I would just omit it to keep things to the point

PR #75, opened 2024-12-12
Worldcereal crop extent resampled

VincentVerelst:
Updated the worldcereal_crop_extent UDP to include a `projection` parameter, which allows the user to set the projection of the final results.

### DIFF
--- a/benchmark_scenarios/worldcereal_crop_extent.json
+++ b/benchmark_scenarios/worldcereal_crop_extent.json
@@ -36,9 +36,9 @@
       ]
     },
     "reference_data": {
-      "cropland-raw_2018-05-01_2019-04-30.tif": "https://s3.waw3-1.cloudferro.com/apex-benchmarks/worldcereal_crop_extent/cropland-raw_2018-05-01_2019-04-30.tif",
-      "job-results.json": "https://s3.waw3-1.cloudferro.com/apex-benchmarks/worldcereal_crop_extent/job-results.json",
-      "openEO.tif": "https://s3.waw3-1.cloudferro.com/apex-benchmarks/worldcereal_crop_extent/openEO.tif"
+      "worldcereal-cropland-extent.tif": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-benchmarks/worldcereal_crop_extent/worldcereal-cropland-extent.tif",
+      "job-results.json": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-benchmarks/worldcereal_crop_extent/job-results.json",
+      "openEO.tif": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-benchmarks/worldcereal_crop_extent/openEO.tif"
     },
     "reference_options": {
       "atol": 1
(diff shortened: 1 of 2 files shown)


VincentVerelst:
@jdries , the four failing unit tests are due to 4 application packages in the algorithm_catalog which are out of scope of this PR

openeo_udp/worldcereal_crop_extent.json
```diff
@@ -1291,27 +1320,13 @@
       }
     },
     {
-      "name": "orbit_state",
-      "description": "The orbit state of the Sentinel-1 data",
+      "name": "projection",
+      "description": "The EPSG code of the output projection",
```
jdries, line 1324: best to specify that UTM is the default, users should know that they do not need to set this value.

openeo_udp/worldcereal_crop_extent.json
```diff
@@ -1291,27 +1320,13 @@
...
-          "DESCENDING"
-        ]
+        "type": "string"
       },
-      "default": "DESCENDING",
+      "default": null,
       "optional": true
     }
-  ],
-  "minimal_job_options": {
```
jdries, line 1307: does it work without job options?

benchmark_scenarios/worldcereal_crop_extent.json
```diff
@@ -36,12 +36,12 @@
...
-      "cropland-raw_2018-05-01_2019-04-30.tif": "https://s3.waw3-1.cloudferro.com/apex-benchmarks/worldcereal_crop_extent/cropland-raw_2018-05-01_2019-04-30.tif",
-      "job-results.json": "https://s3.waw3-1.cloudferro.com/apex-benchmarks/worldcereal_crop_extent/job-results.json",
-      "openEO.tif": "https://s3.waw3-1.cloudferro.com/apex-benchmarks/worldcereal_crop_extent/openEO.tif"
+      "worldcereal-cropland-extent.tif": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-benchmarks/worldcereal_crop_extent/worldcereal-cropland-extent.tif",
+      "job-results.json": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-benchmarks/worldcereal_crop_extent/job-results.json",
+      "worldcereal-cropland-extent-postprocessed.tif": "https://s3.waw3-1.cloudferro.com/swift/v1/APEx-benchmarks/worldcereal_crop_extent/worldcereal-cropland-extent-postprocessed.tif"
     },
     "reference_options": {
-      "atol": 1
+      "atol": 10
```
jdries, line 44: isn't this a rather high tolerance for a discrete classification?

jdries (changes requested):

openeo_udp/worldcereal_crop_extent.json
VincentVerelst: Currently, the only way it works, is to specify job options after calling the UDP

benchmark_scenarios/worldcereal_crop_extent.json
VincentVerelst: Yes, so here we have a bit of a problem:
- The benchmarks compare all bands of the geotiff at once with one single tolerance level
- For the actual classification (0 or 1) this tolerance indeed doesn't make much sense, but also the classification probabilities are included in the geotiffs. These seemed to vary quite a lot, causing the benchmark to always fail with an absolute tolerance of 1

PR #80, opened 2025-01-09
Populating Algorithms: Mogpr from Fusets 

Pratichhya:
The service is also available at https://marketplace-portal.dataspace.copernicus.eu/catalogue/app-details/17

The modification in the openeo_udp of this repo in comparison to the existing one is:
* The dependencies are passed within the files and need not pass separately in the job_options


Pratichhya:
@soxofaan @HansVRP 
Before changing it as Ready from Draft, could you please provide me with your suggestion  on:

* Here, the input parameter is only datacube, and there is no definition/example on if there is a restriction to passing this as a parameter instead of spatial and temporal extent as done in other cases.
* In the benchmark scenario, the process graph includes multiple nodes, but according to [documentation](https://esa-apex.github.io/apex_documentation/guides/udp_writer_guide.html#defining-a-validation-benchmark-scenario)(correct me if I am wrong) it should only have a single node that points to the namespace.

algorithm_catalog/fusets_mogpr.json
```diff
@@ -0,0 +1,136 @@
...
+        ],
+        "license": "other"
+    },
+    "linkTemplates": [],
+    "links": [
+        {
+            "rel": "openeo-process",
+            "type": "application/json",
+            "title": "openEO Process Definition",
+            "href": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/mogpr_v1/openeo_udp/fusets_mogpr/fusets_mogpr.json"
```
HansVRP, line 117: this branch will probably be deleted afterwards, so make sure to update once merged

openeo_udp/fusets_mogpr/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The spatial extent is limited to a maximum size equal to a Sentinel-2 MGRS tile (100 km x 100 km).
+
+## Configuration & Resource Usage
+
+Run configurations for different ROI/TOI with memory requirements and estimated run durations.
+
+### Synchronous calls
+
+TODO: Replace with actual measurements!!!
```
HansVRP, line 18: please do so

openeo_udp/fusets_mogpr/README.md
```diff
@@ -0,0 +1,51 @@
...
+Compute an integrated timeseries based on multiple inputs.
+For instance, combine Sentinel-2 NDVI with Sentinel-1 RVI into one integrated NDVI.
+
+## Limitations
+
+The spatial extent is limited to a maximum size equal to a Sentinel-2 MGRS tile (100 km x 100 km).
+
+## Configuration & Resource Usage
+
+Run configurations for different ROI/TOI with memory requirements and estimated run durations.
```
HansVRP, line 14: please do so

openeo_udp/fusets_mogpr/README.md
```diff
@@ -0,0 +1,51 @@
...
+| 5 km x 5 km    | 2 minutes    |
+| 10 km x 10 km  | 3 minutes    |
+| 50 km x 50 km  | 9 minutes    |
+
+The maximum duration of a synchronous run is 15 minutes.
+For long running computations, you can use batch jobs.
+
+### Batch jobs
+
+TODO: Replace with actual measurements!!!
```
HansVRP, line 34: please do so

openeo_udp/fusets_mogpr/fusets_mogpr.json
```diff
@@ -0,0 +1,55 @@
+{
+    "process_graph": {
+        "applyneighborhood1": {
+            "process_id": "apply_neighborhood",
+            "arguments": {
+                "data": {
+                    "from_parameter": "input_raster_cube"
```
HansVRP, line 7: here we need to discuss if this fits the current APEx way of working,...

HansVRP (changes requested):

openeo_udp/fusets_mogpr/generate.py
```diff
@@ -0,0 +1,50 @@
...
+from pathlib import Path
+from set_path import load_set_path
+from typing import Union
+
+from openeo import DataCube
+from openeo.api.process import Parameter
+from openeo.processes import ProcessBuilder, apply_neighborhood
+from openeo.rest.udp import build_process_dict
+
+from fusets.openeo import load_mogpr_udf
```
HansVRP, line 11: I do not believe that fusets is part of this environment. @soxofaan what would be the prefered way of working here?

openeo_udp/fusets_mogpr/generate.py
```diff
@@ -0,0 +1,50 @@
...
+from fusets.openeo.services.publish_mogpr import NEIGHBORHOOD_SIZE
```
HansVRP, line 12: does it come with fixed input sizes?

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,78 @@
...
+
+def add_directory_to_sys_path(directory):
+    """
+    Adds a directory to the Python sys.path if it's not already present.
+    """
+    if directory not in sys.path:
+        sys.path.insert(0, directory)
+
+@functools.lru_cache(maxsize=5)
+def setup_dependencies(dependencies_url,DEPENDENCIES_DIR):
```
HansVRP, line 45: @soxofaan should we include file_locking to avoid concurrency issues?

algorithm_catalog/fusets_mogpr.json
```diff
@@ -0,0 +1,136 @@
...
+        "http://www.opengis.net/spec/ogcapi-records-1/1.0/req/record-core"
+    ],
+    "geometry": null,
+    "properties": {
+        "created": "2025-01-093T00:00:00Z",
+        "updated": "2025-01-09T00:00:00Z",
+        "type": "apex_algorithm",
+        "title": "Multi output gaussian process regression",
+        "description": "Integrates timeseries in data cube using multi-output gaussian process regression. The service is designed to enable multi-output regression analysis using Gaussian Process Regression (GPR) on geospatial data. It provides a powerful tool for understanding and predicting spatiotemporal phenomena by filling gaps based on other indicators that are correlated with each other.",
+        "cost_estimate": 12,
```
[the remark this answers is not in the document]
Pratichhya: No this was when updating the job_option. As mentioned in the documentation, 'executor-memory': '7g' was set, however I used:
"executor-memory": "1G",
    "executor-memoryOverhead": "500m",
    "python-memory": "3G"

openeo_udp/fusets_mogpr/generate.py
Pratichhya: Yes seems like the size is already defined within FuseTS

algorithm_catalog/fusets_mogpr.json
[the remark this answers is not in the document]
HansVRP: Okay, so did this influence the cost estimate?

Also did you use the same memory settings in your benchmark?

openeo_udp/fusets_mogpr/generate.py
HansVRP: This is also different from the standard way of working, could you take a look into the code to investigate what those standards are?

openeo_udp/fusets_mogpr/README.md
Pratichhya: It fails due to Read_time out (minimum 13/15 mins)

soxofaan:
> Here, the input parameter is only datacube, and there is no definition/example on if there is a restriction to passing this as a parameter instead of spatial and temporal extent as done in other cases.

I'm not sure I understand what your are asking. Is this about raising the issue of lack of documentation? Or just if it is ok to use a single data cube parameter instead of extent parameters?

soxofaan:
> In the benchmark scenario, the process graph includes multiple nodes, but according to [documentation](https://esa-apex.github.io/apex_documentation/guides/udp_writer_guide.html#defining-a-validation-benchmark-scenario)(correct me if I am wrong) it should only have a single node that points to the namespace.

in the docs I just see "typically a single node":

> process graph will typically just contain a single node

so it does not say "it should". At first sight, I think it's ok to have more than one nodes in the benchmark process graph

openeo_udp/fusets_mogpr/set_path.py
soxofaan: you certainly risk concurrency problems here. However it's not trivial here, because you need a locking mechanism that works across multiple executor. I guess solving this properly is a bit out of scope of this PR

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,78 @@
...
+
+from openeo.udf import inspect
+
+
+# Example constants for demonstration
+DEPENDENCIES_DIR1 = 'venv'
+DEPENDENCIES_DIR2 = 'venv_static'
+
+DEPENDENCIES_URL1 = "https://artifactory.vgt.vito.be:443/artifactory/auxdata-public/ai4food/fusets_venv.zip"
+DEPENDENCIES_URL2 = "https://artifactory.vgt.vito.be:443/artifactory/auxdata-public/ai4food/fusets.zip"
```
soxofaan, line 16: I'm not sure if it is very valuable to define these as constants here. Each value is only used once, so you could just use these values directly in the `setup_dependencies()` calls

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def load_set_path() -> str:
+    """
+    loads path setup functions 
+    @return:
+    """
+    import os
+
+    return Path(os.path.realpath(__file__)).read_text()
```
soxofaan, line 78: just curious: why do you need the `os.path.realpath` here (and the embedded `import os`)?. Wouldn't just `Path(__file__).read_text()` work fine?

Also, it seems a bit overkill at the moment to define this `load_set_path()` function here. When you call it from generate.py, you can just do directly `Path("set_path.py").read_text()` which does the same in less code (overall) and requires less clicking around to understand what's happning

soxofaan (commented):
a couple of notes

Pratichhya:
> > Here, the input parameter is only datacube, and there is no definition/example on if there is a restriction to passing this as a parameter instead of spatial and temporal extent as done in other cases.
> 
> I'm not sure I understand what your are asking. Is this about raising the issue of lack of documentation? 

Could be said such 👀  but No No not exactly. It is a concern on if there is standard way to do as for spatial_extent and temporal extent in terms of naming

> Or just if it is ok to use a single data cube parameter instead of extent parameters?

Yes, this one esp, because I saw no example of doing such

openeo_udp/fusets_mogpr/set_path.py
Pratichhya: > just curious: why do you need the `os.path.realpath` here (and the embedded `import os`)?. Wouldn't just `Path(__file__).read_text()` work fine?

I simply fetched the solution from the fusets's load_mogpr_udf itself to make it working 😬 

> Also, it seems a bit overkill at the moment to define this `load_set_path()` function here. When you call it from generate.py, you can just do directly `Path("set_path.py").read_text()` which does the same in less code (overall) and requires less clicking around to understand what's happning

This indeed is a simpler and nice solution and works exactly the same, thank you so much I updated the generate.py and set_path.py with the suggested solution.

openeo_udp/fusets_mogpr/set_path.py
Pratichhya: updated the values directly in the `setup_dependencies`

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,78 @@
...
+        file.write(response.content)
+
+
+def extract_zip(zip_path, extract_to):
+    """
+    Extracts a zip file from zip_path to the specified extract_to directory.
+    """
+    with zipfile.ZipFile(zip_path, "r") as zip_ref:
+        zip_ref.extractall(extract_to)
+    os.remove(zip_path)  # Clean up the zip file after extraction
```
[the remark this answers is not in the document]
Pratichhya: replaced the os.remove(zip_path) within the setup_dependencies function

benchmark_scenarios/fusets_mogpr.json
```diff
@@ -0,0 +1,181 @@
...
+                                    5.170012098271149,
+                                    51.25062964728295
+                                ]
+                            ]
+                        ],
+                        "type": "Polygon"
+                    },
+                    "temporal_extent": [
+                        "2022-05-01",
+                        "2023-07-31"
```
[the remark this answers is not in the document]
Pratichhya: No that was not intentional and thanks for pointing as it is one of the cause of high credits 😅 
Updated the scenario to use only a month.

openeo_udp/fusets_mogpr/generate.py
Pratichhya: NEIGHBORHOOD_SIZE used in 32px

https://github.com/Open-EO/FuseTS/blob/main/src/fusets/openeo/services/publish_mogpr.py#L12

Instead of importing the values, I type in 32 in the apply_neighborhood of generate.py

openeo_udp/fusets_mogpr/generate.py
Pratichhya: In addition to generating the UDP it is also when publishing and running the UDP.
Updated the requirements.txt

openeo_udp/fusets_mogpr/fusets_mogpr.json
Pratichhya: > I'd suggest to just use `data` as parameter name (like in the majority of raster cube processes) if you have a single data cube as input parameter

Updated the parameter name from "input_raster_cube" to "data".

But, @JanssenBrm @jdries instead of requesting for datacube: 
* suggesting to do the load collection  
* masking
* call udp for mogpr  
* aggregate for timeseries.

As shown in the usage example here: https://marketplace-portal.dataspace.copernicus.eu/catalogue/app-details/17

Is there a reason for why are we not requesting only for spatial and temporal extent and returning back the filled time-series in this service?

openeo_udp/fusets_mogpr/README.md
Pratichhya: updated

openeo_udp/fusets_mogpr/README.md
Pratichhya: remove the measurements section in the readme

algorithm_catalog/fusets_mogpr.json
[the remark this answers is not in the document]
Pratichhya: not much: by 4/5 credits.
Usually in my setting it is around 15 credits but with the executor-memory: 7g it is 19credits.

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def add_directory_to_sys_path(directory):
+    """
+    Adds a directory to the Python sys.path if it's not already present.
+    """
+    if directory not in sys.path:
+        sys.path.insert(0, directory)
```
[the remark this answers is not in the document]
soxofaan: prepend = insert at index 0

prepending to sys.path is usually an antipattern (because of security and stability risks), and we should not implicitly promote that by copying that pattern all over the place.

> because there was a conflict in modules already existing to that of extracted

that's what I mean: trying to fix this problem by prepending is like shooting a mosquito with a bazooka: you risk to break a lot more than what you intend to.

>  it couldn't find the fusets as done with append as seen in "j-2501141413334eaab23c071c8db34078"

In these error logs I see `ModuleNotFoundError: No module named 'fusets'` . I don't really get how prepending instead of appending would fix that.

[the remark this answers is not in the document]
HansVRP: @Pratichhya  I can assist you on digging deeper into this issue; 

Perhaps, there is another file added to system path with the same name?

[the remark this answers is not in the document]
Pratichhya: This could be because the dependency includes not only fusets but also many other packages with a specific version. Not sure if they were affected.

openeo_udp/fusets_mogpr/fusets_mogpr.json
Pratichhya: Discussed the reason in person and the upgraded version of this process "MOGPR" is "MOGPR_s1_s2". So, I will update this process in the same PR addressing all the suggested comments since the content for both will be almost the same.

openeo_udp/fusets_mogpr/generate.py
HansVRP: now I better understand, this would not need to be added to the requirements I believe

openeo_udp/fusets_mogpr/set_path.py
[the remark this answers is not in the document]
HansVRP: Just tested, the core reason is 

DEPENDENCIES_URL1 = "https://artifactory.vgt.vito.be:443/artifactory/auxdata-public/ai4food/fusets_venv.zip"

It contains a fusets/openeo folder which does not contain load_mogpr_udf. 

When both are appended the backend looks in both and does not find the udf back in the first folder it looks in. 
The fact that it did work with sys.path.insert is probably a coincidence due to the ordering of adding the dependency folders.

I would propose to discuss with the developer to update the fusets_venv. zip by either including the latest fusets from fusets.zip. This is also cleaner since you will only need to add in 1 dependency folder

openeo_udp/fusets_mogpr/README.md
```diff
@@ -0,0 +1,39 @@
+# Sentinel-1 and Sentinel-2 data fusion through Multi-output Gaussian process regression (MOGPR)
+
+This service is designed to enable multi-output regression analysis using Gaussian Process Regression (GPR) on geospatial data. It provides a powerful tool for understanding and predicting spatiotemporal phenomena by filling gaps based on other correlated indicators. This service focusses on the fusion of Sentinel-1 and Sentinel-2 data, allowing the user to select one of the predefined data sources.
```
SergeCroise, line 3:
```suggestion
This service is designed to enable multi-output regression analysis using Gaussian Process Regression (GPR) on geospatial data. It provides a powerful tool for understanding and predicting spatiotemporal phenomena by filling gaps based on other correlated indicators. This service focuses on the fusion of Sentinel-1 and Sentinel-2 data, allowing the user to select one of the predefined data sources.
```

SergeCroise, line 3: **focuses** or _focusses_ ?

Pratichhya:
Finally ready, I guess

qa/unittests/requirements.txt
```diff
@@ -3,3 +3,4 @@ git+https://github.com/ESA-APEx/esa-apex-toolbox-python.git@main
 pytest>=8.2.0
 moto[s3, server]>=5.0.13
 dirty-equals>=0.8.0
+fusets>=2.0.1
```
soxofaan, line 6: No, it should not be added here.
fusets is not a dependency for the unit tests.

I meant to add a requirements.txt file to the openeo_udp/fusets_mogpr/folder

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Extract the zip file to the temporary directory
+    with zipfile.ZipFile(zip_path, "r") as zip_ref:
+        zip_ref.extractall(temp_dir)
+
+    return temp_dir
+
+def move_top_level_folder_to_destination(temp_dir, destination_dir):
+    """
+    Moves the first top-level folder from the temporary directory to the destination directory.
```
soxofaan, line 36: doc states "first top-level folder", but in implementation I see a for loop, so that means "each top level folder"

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,92 @@
...
+    inspect(message="Move top-level folder to destination")
+    moved_folder = move_top_level_folder_to_destination(extracted_dir, destination_dir)
+
+    # Step 4: Add the folder to sys.path
+    add_to_sys_path(moved_folder)
+    inspect(message="Added to the sys path")
+
+    # Clean up the temporary zip file
+    os.remove(zip_path)
+    shutil.rmtree(temp_dir)  # Remove the temporary extraction folder
```
soxofaan, line 89: Avoid a pattern like
```python
temp_dir = tempfile.mkdtemp()
# stuff
shutil.rmtree(temp_dir) # Remove the temporary extraction folder
```

Instead, use a context manger:
```
with tempfile.TemporaryDirectory() as temp_dir:
    # stuff
```

then the temp dir will be always cleaned up automatically, even if an exception happened during `# stuff`

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,92 @@
...
+    response = requests.get(url, stream=True)
+    with open(path, "wb") as file:
+        file.write(response.content)
+
+def extract_zip_to_temp(zip_path):
+    """
+    Extracts a zip file to a temporary directory.
+    """
+    # Create a temporary directory
+    temp_dir = tempfile.mkdtemp()
```
soxofaan, line 26: Note: this temp dir is not cleaned up

I would do all the download/unzip stuff in the same temp folder instead of creating a separate temp folder for each step

openeo_udp/fusets_mogpr/helpers.py
```diff
@@ -0,0 +1,213 @@
...
+#######################################################################################################################
+def _build_collection_graph(collection, label, callable, reject):
+    """
+    Helper function that will construct an if-else structure using the if_ openEO process. If the value of the
+    collection parameter matches with the given label, the callable is executed. If not the reject function is
+    executed.
+
+    :param collection: openEO collection parameter
+    :param label: String representing the text with which the collection should match
+    :param callable: Function that is executed when the collection matches the label
```
soxofaan, line 128: this is a bit confusing: the value you pass to `callable` (also called "function" in other places of this script) is not a callable/function but a data cube

soxofaan (commented):
some more notes :)

openeo_udp/fusets_mogpr/helpers.py
```diff
@@ -0,0 +1,213 @@
...
+        collections = _build_collection_graph(collection=collection,
+                                              label=option['label'],
+                                              callable=option['function'],
+                                              reject=collections)
+    return collections
+
+
+def load_s2_collection(connection, collection, polygon, date):
+    """
+    Create a S2 input data cube based on the collection selected by the user. This achieved by building an
```
SergeCroise, line 167:
```suggestion
    Create a S2 input data cube based on the collection selected by the user. This is achieved by building an
```

openeo_udp/fusets_mogpr/helpers.py
```diff
@@ -0,0 +1,213 @@
...
+    :param callable: Function that is executed when the collection matches the label
+    :param reject: Function that is executed when the collection does not match the label
+    :return:
+    """
+    return if_(eq(collection, label, case_sensitive=False), callable, reject)
+
+
+def load_s1_collection(connection, collection, polygon, date):
+    """
+    Create a S1 input data cube based on the collection selected by the user. This achieved by building an
```
SergeCroise, line 137:
```suggestion
    Create a S1 input data cube based on the collection selected by the user. This is achieved by building an
```

algorithm_catalog/fusets_mogpr.json
```diff
@@ -0,0 +1,136 @@
...
+    "conformsTo": [
+        "http://www.opengis.net/spec/ogcapi-records-1/1.0/req/record-core"
+    ],
+    "geometry": null,
+    "properties": {
+        "created": "2025-01-09T00:00:00Z",
+        "updated": "2025-01-29T00:00:00Z",
+        "type": "apex_algorithm",
+        "title": "Multi output gaussian process regression",
+        "description": "Integrates timeseries in data cube using multi-output gaussian process regression. The service is designed to enable multi-output regression analysis using Gaussian Process Regression (GPR) on geospatial data. It provides a powerful tool for understanding and predicting spatiotemporal phenomena by filling gaps based on other indicators that are correlated with each other.",
```
SergeCroise, line 13: **time series** or _timeseries_ ?

openeo_udp/fusets_mogpr/generate.py
```diff
@@ -0,0 +1,86 @@
...
+        polygon=polygon,
+        date=date,
+        s1_collection=s1_collection,
+        s2_collection=s2_collection
+    )
+
+    return build_process_dict(
+        process_graph=mogpr,
+        process_id="fusets_mogpr",
+        summary="Integrate S1 and S2 timeseries using multi-output gaussian process regression",
```
SergeCroise, line 72: **time series** or _timeseries_ ?

openeo_udp/fusets_mogpr/fusets_mogpr.json
HansVRP: was this done @Pratichhya ?

openeo_udp/fusets_mogpr/generate.py
HansVRP: if its multiple times used, you can set it as a global

openeo_udp/fusets_mogpr/set_path.py
HansVRP: @Pratichhya pas this sys.path.insert replaced here?

openeo_udp/fusets_mogpr/set_path.py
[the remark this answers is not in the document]
HansVRP: Is this done @Pratichhya

openeo_udp/fusets_mogpr/fusets_mogpr.json
Pratichhya: Yes, I updated the process graph

openeo_udp/fusets_mogpr/generate.py
Pratichhya: No, it was used only for this single function

openeo_udp/fusets_mogpr/set_path.py
Pratichhya: Yes, has been updated

openeo_udp/fusets_mogpr/set_path.py
HansVRP: was this done @Pratichhya?

openeo_udp/fusets_mogpr/set_path.py
HansVRP: Did you manage to get this to work @Pratichhya or should we take a look together?

openeo_udp/fusets_mogpr/set_path.py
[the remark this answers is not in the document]
Pratichhya: Informed

openeo_udp/fusets_mogpr/set_path.py
Pratichhya: I guess so

openeo_udp/fusets_mogpr/set_path.py
Pratichhya: Yes, updated.

openeo_udp/fusets_mogpr/helpers.py
HansVRP: for option in [
        {
            'label': 'grd',
            'function': _load_s1_grd_bands(connection=connection, polygon=polygon, date=date, bands=['VV', 'VH'])
        },
        
       Probably a matter of syntax, here they pass a callable function which creates a cube. 
       
       For me it would be fine to keep things as is since @Pratichhya  is not the original developer. 
       Do you agree @soxofaan?

HansVRP: Probably a matter of syntax, here they pass a callable function which creates a cube. 
       
For me it would be fine to keep things as is since @Pratichhya  is not the original developer. 
Do you agree @soxofaan?

openeo_udp/fusets_mogpr/set_path.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Step 3: Move the first top-level folder (dynamically) to the destination
+        destination_dir = os.getcwd()  # Current working directory
+        inspect(message="Move top-level folder to destination")
+        moved_folder = move_top_level_folder_to_destination(extracted_dir, destination_dir)
+
+        # Step 4: Add the folder to sys.path
+        add_to_sys_path(moved_folder)
+        inspect(message="Added to the sys path") 
+
```
HansVRP, line 80: @Pratichhya I believe here we still need to clean up the temp dir as indicated by stefaan

HansVRP:
@Pratichhya for me it is good to merge, but first rebase to the local main and check that the unit tests are passing

Pratichhya:
> @Pratichhya for me it is good to merge, but first rebase to the local main and check that the unit tests are passing

Rebased and the failure is not with the changes in this branch 😄

openeo_udp/fusets_mogpr/helpers.py
soxofaan: I'm not sure what you mean.   `_load_s1_grd_bands(...)` is a function _call_ , right, not a callable?
And it's results in a data cube, so  you have a dictionary
```
{
    "function": <DataCube Object>
}
which looks weird to me.


But as mentioned elsewhere, don't let all my comments be blockers. I'm fine with accepting things as is, and clean up things later.

Pratichhya:
rebased again and no failure
@soxofaan @HansVRP 
Can I go ahead and merge this PR?

soxofaan:
I'm afraid there is some more work to be done as with #88 the whole folder structure has been changed, so you would have to move some files around in this PR

HansVRP:
indeed, however, it should be simpler now since we have a folder per service!

Issue #81, opened 2025-01-09
Restructure the folders of the algorithm catalogue


JanssenBrm:
The current structure of the algorithm could be optimized to simplify the management of the integrated algorithms and their details. 

I propose to apply the following changes to the structure. For existing folders that are not in this list, I don't have any opinion. 

```
.
├── service_catalogue/
│   └── <organization>/
│       └── <backend>/
│           └── <service_ogc_api_record>.json
├── benchmark_scenarios/
│   └── <organization>/
│       └── <scenario>.json
└── openeo_udp/
    └── <organization>/
        └── <service>/
            ├── <description>.md
            └── <other relevant files>
````

With the following main changes:

* Rename `algorithm_catalog` to `service_catalogue` to avoid any confusion - we are promoting deployed services and not the algorithms themselves
* Create an organization (e.g. VITO, Terradue, ...) and backend-based (CDSE, GEP, ...) subfolder structure to provide a bit more structure 
* Moving the description files for openEO UDPs inside the `openeo_udp` folder instead of storing them on the level of the OGC API Record. 

In general,  the `service_catalogue` folder should only contain the OGC API Records. From these records, any external tool should be able to get all relevant information.

JanssenBrm:
@HansVRP  - Tagging you here to also get your opinion on the matter.

JanssenBrm:
Also tagging @HerveCaumont / @p3dr0  - feel free to share your suggestions/objections.

HansVRP:
For an external user having a folder per surface might be easier; what is the core reason we split their contributions up?

JanssenBrm:
> For an external user having a folder per surface might be easier; what is the core reason we split their contributions up?

Thank you for your feedback Hans, can you maybe give an example of having a folder per service? Is this within the `service_catalogue`, `benchmark_scenarios`, `openeo_udp`or all of them? 

If splitting up the contributions is referring to the `organization` subfolders, this was suggested to enable organisations to have a clear overview of what has been integrated by them. For now, we only have VITO and Terradue, but any organisation could ingest a record into the catalogue.

HansVRP:
something like:
.
├── <organization 1>/
          └── <backend>/
│                  └── <service_ogc_api_record>.json
│       └── <service 1>
│               └── benchmark_scenarios/
│               └── openeo_udp/
│       └── <service 2>
│               └── benchmark_scenarios/
│               └── openeo_udp/

├── <organization 2>/
          └── <backend>/
[...]

Issue #16, opened 2024-07-01
Rethink or be careful with usage of term "UDP"


soxofaan:
In the context of this project (and outside of it too), we often use the term "UDP" (user-defined process), e.g. in tickets, docs, code etc for reusable openEO processes. 

But I think we have to be careful here with this term to not cause [more confusion](https://github.com/Open-EO/openeo-api/issues/353) than necessary.
Processes defined (and propagated) here are defined by algorithm providers, not (end) users.

So I would go for  more generic terms like
- openEO process definition
- openEO process

soxofaan:
related:
- https://github.com/Open-EO/openeo-api/issues/540

jdries:
I started applying this in latest writing.
To make a transition, I'm going for "openEO process (UDP)".
That's because an important part of our existing audience is familiar with and expects to see the UDP term.

soxofaan:
> That's because an important part of our existing audience is familiar with and expects to see the UDP term.

indeed we should certainly mention and clarify the link with the commonly used term "UDP"

PR #88, opened 2025-01-23
chore(#81): updated structure

JanssenBrm:
This a PR to follow up the discussions in #81.


JanssenBrm:
It probably requires a bit more work to ensure all of the references are still working.

HansVRP:
Took a diagonal view and the structure looks much cleaner. I noticed currently do not consistently have the records folder to contain all png images ect

HansVRP:
Will provide a clear list soon from vito's side

soxofaan:
I also prefer the proposed structure, it will make overview and maintenance easier I think.

One note: while we do this restructuring I think we should also eliminate the usage of the term "UDP", e.g. as used in folder name "openeo_udp" here. As raised in https://github.com/ESA-APEx/apex_algorithms/issues/16, the "U" in "UDP" (which stands for "user") is not ideal to use in the context of APEx I think because the algorithms are not defined by the user, but algorithm providers. I the proposed folder restructuring, I think we can just use folder name "openeo"

JanssenBrm:
Unit tests are failing because they are actually looking for files in the new folder structure on the main branch.

HansVRP:
apologies, then all is fine for me!

PR #99, opened 2025-02-18
Alignment on the relationship types

JanssenBrm:
This PR contains t he following changes:

- Introduction of a record schema - this schema is built using the [EarthCODE schema](https://github.com/ESA-EarthCODE/open-science-catalog-validation/blob/main/schemas/records.json)  as a base, modified with the sections that are dedicated for APEx
- Validation of the record through a unit test
- Update of existing records to ensure that they are valid.

### DIFF
--- a/algorithm_catalog/terradue/gep_bas/records/gep_bas.json
+++ b/algorithm_catalog/terradue/gep_bas/records/gep_bas.json
@@ -40,6 +40,7 @@
         "links": [
           {
             "href": "https://geohazards-tep.eu",
+            "title": "Geohazards Exploitation Platform",
             "rel": "about",
             "type": "text/html"
           }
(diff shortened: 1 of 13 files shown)


soxofaan (approved):
very nice and much needed apparently

soxofaan:
> this schema is built using the [EarthCODE schema](https://github.com/ESA-EarthCODE/open-science-catalog-validation/blob/main/schemas/records.json) as a base, modified with the sections that are dedicated for APEx

I wonder how we can keep track of this relationship. E.g. when the EarthCODE schema changes a bit?

Should we also validate the records against the vanilla EarthCODE schema?

JanssenBrm:
> > this schema is built using the [EarthCODE schema](https://github.com/ESA-EarthCODE/open-science-catalog-validation/blob/main/schemas/records.json) as a base, modified with the sections that are dedicated for APEx
> 
> I wonder how we can keep track of this relationship. E.g. when the EarthCODE schema changes a bit?
> 
> Should we also validate the records against the vanilla EarthCODE schema?

If possible this would be useful, but they will also require some specific open science metadata. I don't have enough experience with the tools to see if we can "exclude" certain rules.

PR #105, opened 2025-02-27
updated misplaced udf

Pratichhya:
When revisiting the UDP, I noticed that the UDF node was missing, probably misplaced in some commit. Thus re-updating the udp json.

https://github.com/ESA-APEx/apex_algorithms/tree/main/algorithm_catalog/vito/biopar/openeo_udp

misplaced in commit: https://github.com/ESA-APEx/apex_algorithms/commit/504895642a7b3a408a2156a2c81bc1e19a27f76d


algorithm_catalog/vito/biopar/benchmark_scenarios/biopar.json
```diff
@@ -2,46 +2,51 @@
...
-                "result": true
+                        ],
+                        "type": "Polygon"
+                    },
+                    "temporal_extent": [
+                        "2020-06-27",
+                        "2020-07-27"
+                    ],
+                    "result": true
+                }
```
HansVRP, line 49: Why doe we need to change the benchmark condition? Were the input wrong or?

algorithm_catalog/vito/biopar/openeo_udp/biopar.json
```diff
@@ -1,195 +1,237 @@
...
+                "data": {
+                    "from_node": "reducedimension1"
+                },
+                "label": {
+                    "from_parameter": "biopar_type"
+                },
+                "name": "bands",
+                "type": "bands"
+            },
+            "result": true
```
HansVRP, line 99: oh, so the UDF wasn't called in the pg?

algorithm_catalog/vito/biopar/benchmark_scenarios/biopar.json
Pratichhya: No, it was correct. But to confirm I re-tested with latest udp json for a different biopar type and thought of updating it too

algorithm_catalog/vito/biopar/openeo_udp/biopar.json
Pratichhya: yes, seems like I mistakely accept the incorrect changes while commiting 😢

HansVRP: do we want to have the generate.py, in this folder as well?

Pratichhya: not sure, if needed I can include it here but I was worried on replicating the content

HansVRP: just to be sure that we are on the same line: My question was: 

Does the openeo_udp folder contain the generate.py file to this UDP? Or does it not make sense to add it here?
Adding it in the folder would help if future improvements need to be done, then we can easily rebuild the UDP.

How would it replicate content?

Pratichhya: Ahh no, it is not there yet. The changes were usually made in cdse marketplace repo so didn't bring the generate.py here.

But is just two files so, if it would be helpful I can update it here

Pratichhya: @HansVRP the generate file is also included now within the openeo_udp folder as for others

Pratichhya:
Rebased to resolve conflict

HansVRP:
good to merge then @Pratichhya ?

PR #135, opened 2025-03-24
fix

HansVRP:
fixed the unit tests, for the windmill detection I added support for openeo.cloud based UDPs. 

Would you rather switch to the federated endpoint in the UDP itself?


jdries:
Merged!
In general, we're promoting use of federated endpoints everywhere, in an attempt to reduce endpoint confusion.

PR #219, opened 2025-08-08
Hv add wac

HansVRP:
pipeline for worldagrocommodities


JanssenBrm:
Is this a service that could be executed through the web editor? If so, it would be good to explicitly add a `webapp`  href to the links section. Similar to https://github.com/ESA-APEx/apex_algorithms/blob/main/algorithm_catalog/vito/biopar/records/biopar.json#L123-L128

JanssenBrm:
The same if there is any example outputs that we could share (https://github.com/ESA-APEx/apex_algorithms/blob/main/algorithm_catalog/vito/biopar/records/biopar.json#L117-L122)

JanssenBrm (approved):

HansVRP:
@JanssenBrm  did an update to the latest UDP version, and added the private property

PR #252, opened 2025-10-02
Issue81_bapCompositeUpdate

tho-hole:
Updated the .json file of BAP_composite UPD to reflect changes made in Issue81 of OpenEO community_examples.


tho-hole:
You can find a simple testcode, based on the bap_composite example in a seperate branch:
https://github.com/ESA-APEx/apex_algorithms/tree/Issue81_bapCompositeUpdate_testcodes

tho-hole:
Previous UDP could succesfully run with a negative cloud_cover (job j-2510011253474fdc9708ec23673f97f5). 
I've added a min-max value checker on cloud_cover, which makes the process fail but only on the server side. I don't see a solution to stop the process before job submission.

PR #271, opened 2025-11-12
added the missing helper file

Pratichhya:
We noticed that a file was missing, likely during the migration process.
Nevertheless, it was already referenced in the process graph, so it does not change the graph.

### DIFF
--- a/algorithm_catalog/vito/fusets_mogpr/openeo_udp/helpers.py
+++ b/algorithm_catalog/vito/fusets_mogpr/openeo_udp/helpers.py
@@ -1,0 +1,213 @@
+"""
+Helper functions to load the different input data cubes for the FUSETS MoGPR algorithm.
+"""
+from openeo.processes import process, if_, eq
+
+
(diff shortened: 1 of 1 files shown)


algorithm_catalog/vito/fusets_mogpr/openeo_udp/helpers.py
```diff
@@ -0,0 +1,213 @@
...
+    :param date: Time of interest
+    :param bands: Bands to load
+    :return:
+    """
+    s1_grd = connection.load_collection('SENTINEL1_GRD',
+                                        spatial_extent=polygon,
+                                        temporal_extent=date,
+                                        bands=bands)
+    s1_grd = s1_grd.sar_backscatter(coefficient='sigma0-ellipsoid')
+    s1_grd = s1_grd.rename_labels(dimension="bands", target=bands)
```
HansVRP: why is this rename_labels required?

Issue #264, opened 2025-10-27
Port `parcel_delineation` UDFs to python 3.11


soxofaan:
per https://github.com/eu-cdse/openeo-cdse-infra/issues/767, the benchmarks `parcel_delineation` was pinned down to python 3.8 with  https://github.com/ESA-APEx/apex_algorithms/commit/52c029cc843caedc7705e0a8d39eee4a143416fe + https://github.com/ESA-APEx/apex_algorithms/commit/1e3f985ec46c62086f3e4406a0f9f572c43f5865 + https://github.com/ESA-APEx/apex_algorithms/commit/33f47803591b47945912db5d21c20fe1429760dd

These UDFs should be ported to python 3.11

PR #273, opened 2025-11-12
Parceldelineation python311

manugv:
I have updated the files and tested to run on python 311. However, the results are different. This can be due to update on S2 data, and scikit-image libraries. Also the files [.py] are synced with the community examples. So the `reference_data` in `benchmark_scenarios/parcel_delineation.json` need to be updated.


manugv:
https://github.com/ESA-APEx/apex_algorithms/issues/264

- [ ] Still `reference_data` needs to be updated.

soxofaan:
> I would prefer if we merge this and wait for it to produce actual data. I

:+1: 

you can merge yourself I assume?

PR #299, opened 2025-12-23
Forestfire scenario failure

Pratichhya:
Updated the scikit-learn dependency in the UDF itself.


Pratichhya:
@VictorVerhaert kind reminder

PR #306, opened 2026-01-16
 Add Brockmann Consult GmbH as provider and the EFAST UDP

Hneuschmidt:
Follow-up to  #295 and  #303

### DIFF
--- a/algorithm_catalog/bc/efast/records/efast.json
+++ b/algorithm_catalog/bc/efast/records/efast.json
@@ -81,16 +81,22 @@
   "linkTemplates": [],
   "links": [
     {
-        "rel": "related",
+        "rel": "code",
         "type": "text/html",
         "title": "Source code repository used to generate the process graph",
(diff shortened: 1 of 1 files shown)


Hneuschmidt:
I have updated the rel types of the newly added links.
For the example notebook, I have found only one example in the repository, which uses rel type "notebook": https://github.com/ESA-APEx/apex_algorithms/blob/4e4372d95dbccc22e7c72edaf6962a36bb8581d5/algorithm_catalog/vito/parcel_delineation/records/parcel_delineation.json#L145

I have made the same choices as in this example, using

```json
"type": "text/html"
```

and a link to the raw notebook json file.

Please let me know if it would be preferable to use the mime type `application/x-ipynb+json`, as used by the jupyter project itself \[1\], or to link to the github page containing the notebook, which renders it in a suitable way for viewing in a browser.

\[1\]: [https://docs.jupyter.org/en/latest/reference/mimetype.html](https://docs.jupyter.org/en/latest/reference/mimetype.html)

Issue #336, opened 2026-01-29
PPI Process 'ppi' parameter 'geometry' is required


EmileSonneveld:
`tests.test_benchmarks.test_run_benchmark[bap_composite_stac308]`
`openeo.rest.OpenEoApiError: [400] ProcessParameterRequired: Process 'ppi' parameter 'geometry' is required. (ref: r-2601281258014aa392687299be0fca64)`
https://jenkins.vgt.vito.be/job/openEO/job/openeo-apex-benchmarks/417/testReport/tests/test_benchmarks/test_run_benchmark_ppi__2/
Probably `geom` needs to be renamed

PR #340, opened 2026-01-29
geom->geometry

EmileSonneveld:
https://github.com/ESA-APEx/apex_algorithms/issues/336
@codex review

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -1,6 +1,6 @@
 name: "Unit Tests"
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   unittests:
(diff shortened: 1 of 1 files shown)


jdries:
There always has to be reference data, you can use (very) lenient tolerances if you want to relax the comparison, but we need to know what it is supposed to produce.

soxofaan:
benchmarks without reference data should break, otherwise the test utilities won't upload actual results to s3 to use be used as reference data in subsequent runs

skipping these benchmarks will break the intended workflow of the benchmarks

EmileSonneveld:
@rahmandawibowo-vito [notify-catalogue-web](https://github.com/ESA-APEx/apex_algorithms/actions/runs/21516569061/job/61997138059?pr=340) fails, do you know how to fix it? I retried a few times

PR #407, opened 2026-03-14
updated to larger extent

Pratichhya:
change in timesteps in netCDF output was observed with the test results. 
A similar issue was encountered in the previous test as well, where 1 timestamp was missing and this time it is 2 timestep. 

Furthermore, the test area was very small and small change in the output results as big difference. Therefore updated a refernce image for larger area with same temporal extent.

### DIFF
--- a/algorithm_catalog/vito/peakvalley/benchmark_scenarios/peakvalley.json
+++ b/algorithm_catalog/vito/peakvalley/benchmark_scenarios/peakvalley.json
@@ -11,11 +11,10 @@
                     "recovery_ratio": 1,
                     "slope_threshold": 0.007,
                     "spatial_extent": {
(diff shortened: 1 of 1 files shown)


Pratichhya:
@JeroenVerstraelen I forgot about this PR. Is the bechmark for Peakvalley still failing if so, this PR updated metadata with larger extent for reference data.

Or would you recommend closing this PR for now?

JeroenVerstraelen:
@Pratichhya this benchmark is no longer failing:
https://jenkins.vgt.vito.be/job/openEO/job/openeo-apex-benchmarks/535/testReport/tests/test_benchmarks/test_run_benchmark_peakvalley_/

If a larger extent is not required then closing this PR might be best.

PR #415, opened 2026-03-18
OBSGESSION LAI

### LINKED ISSUE #487: Integrate LAI-EVI UDP in the APEx Algorithm Catalogue
The resulting EVI-LAI pipeline is published to the APEx Algorithm Catalogue, where it can be registered, versioned, and made discoverable for execution on compatible openEO backends. This publication step ensures that workflows are reusable, traceable, and aligned with APEx onboarding procedures for EO services.

JanssenBrm:


DeRooBert:
@JanssenBrm it seems that a commit was not yet in dev. I asked for a pull request. After the pull request the problem should be solved. One extra remark is that it will be much easier to just adapt the json and not rerun it since there are plenty of manual stuff todo. Since the openeo has checks as well that do not go well with parameters.

Pratichhya:
Description updated for the process. 

Currently revisiting the generate progress graph workflow, thus marking as draft PR.

Pratichhya:
@HansVRP seems like I cannot request @JanssenBrm for his own PR review 😄.

PS: waiting for thumnail from Eline

https://esa-apex.github.io/apex-algorithms-catalogue-web/pr-preview/pr-415/apps/udp_obsgession_w23_lai/

Pratichhya:
Completed all the suggested changes. Let me know if additional changes are needed.

HansVRP:
there seem to be some unit tests failing with your entry: 

FAILED tests/test_scenarios.py::test_lint_scenario[udp_obsgession_w23_lai] - ValueError: Links should not point to ephemeral feature branches: found 'obsgession_lai' in 'https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/obsgession_lai/algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_obsgession_w23_lai.json'

algorithm_catalog/obsgession/udp_obsgession_w23_lai/records/udp_obsgession_w23_lai.json
```diff
@@ -0,0 +1,119 @@
...
+    "languages": [
+      {
+        "code": "en-US",
+        "name": "English (United States)"
+      }
+    ],
+    "contacts": [
+      {
+        "name": "Marcel Buchhorn",
+        "position": "NCA team",
```
HansVRP: Senior Researcher, or Prime researcher, ...

algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_generation.py
```diff
@@ -0,0 +1,223 @@
...
+            )
+    param_binning_period= Parameter.string(
+        name="binning_period",
+        default='month',
+        description="The temporal binning period. Please have a look at openeo documentation of the process "
+                    "aggregate_temporal_period for more information",
+        values=['day', 'week', 'dekad','month', 'season', 'year']
+    )
+
+    param_temp_aggregator = Parameter.string(
```
HansVRP, line 102: why does this need to be a parameter?

algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_generation.py
```diff
@@ -0,0 +1,223 @@
...
+        name="temp_aggregator",
+        description="The temporal aggregation function. Please have a look at openeo documentation of the process "
+                    "aggregate_temporal_period for more information",
+        values=["min", "max", "mean", "median"],
+        default="mean"
+    )
+    param_epsg = Parameter.number(
```
HansVRP, line 109: why does this need to be a parameter?

HansVRP: very dangerous to do so since there is a coupling to the resolution

algorithm_catalog/obsgession/udp_obsgession_w23_lai/records/udp_obsgession_w23_lai.json
Pratichhya: I mentioned Reseracher alone but he suggested this instead.
I will update it to Senior Resercher, I assume that should be fine.

algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_generation.py
Pratichhya: I confirmed this with Marcel and he mentioned that they noticed some user passing different type of aggregator. And since the algorithm is capable of providing the solution based on that. 

So I thought to keep the default as mean unless user wish to change it.

algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_generation.py
Pratichhya: Same reason as above.  If I understood well, outside openEO they included a package that adapts irrelevant to which user request but not with openEO. 

To avoid frequent intervention to this too, the default is set to 3035.

Pratichhya: so the workflow will be simple as only passing temporal and spatial extent, unless user themselves play around with other params

HansVRP:
@Pratichhya there is still a failing unit test; you need to use the following raw link and not one which includes 'ref/head, ...'

https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/obsgession_lai/algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_obsgession_w23_lai.json

Pratichhya:
Thank you @HansVRP for the guidance. It was still failing with your suggestion, so instead I included the reference to the latest commit. I assume that works too 😅

algorithm_catalog/obsgession/udp_obsgession_w23_lai/openeo_udp/udp_generation.py
mbuchhorn: > very dangerous to do so since there is a coupling to the resolution

I agree with Hans on that point. Normally we have a check that when the user requests a projected coordinate system that the resolution is above 10 AND when the requested crs is a geographic one that the resolution is under 1. 
But for that UDP we had removed it.

Anyhow.... I will talk to the community if we want to limit this service to pan-EU.... then we can omit the epsg parameter! But we need the resolution parameter! Next to a default value of 10m, we can add a check that when the parameter value is <10 or >300 then it is reset to 10 :) That should be easy with a IF_ clause, or?

PR #442, opened 2026-03-31
Allow httpS://www.opengis.net/spec/ogcapi-records-1/1.0/req/record-core

EmileSonneveld:
The http version gives security warnings in my editor, but when I fixed it, it broke validation.
This will allow both http and https to be used, allowing a grace period to transition to https.


JanssenBrm:
Thank you @EmileSonneveld. Maybe better to only allow HTTPS?

EmileSonneveld:
That would be best, but I fear that there are some MRs that might break due to this.
This would be the non-interrupting way

JanssenBrm:
The value is only used for validation and should not impact any existing functionalities, so it would be better to just update the existing records accordingly

EmileSonneveld:
@copilot, review without making code changes.

EmileSonneveld:
`TestGithubAlgorithmRepository::test_get_algorithm` refers to the main branch, so also here, I'll need to do a small cleanup after merge

PR #454, opened 2026-04-07
Statistic udp

HansVRP:
A new entry for a general statistic UDP; 

calculates the mean std min max Q10 Q50 and Q90 per band. 

users can be flexible in terms of temporal period; they can calculate the statistics for the full time range, or get a time series by selecting a period.


HansVRP:
output will look look like: 

image

with a row per timestep

JanssenBrm (approved):

PR #457, opened 2026-04-09
statistics_udp: fixed refs to point to main

JanssenBrm:


HansVRP:
@soxofaan I do not think the current testing allows for comparing output csv files; do you know if there is already an issue on this?

soxofaan:
no CSV is not covered yet

there is an open ticket for vector cube comparison at https://github.com/Open-EO/openeo-python-client/issues/862  however, which is probably related

PR #466, opened 2026-04-15
better representation github issues

HansVRP:
proposal to make the github tickets more fractional; by having a label related to the stage of failure we can more easily track down which jobs (failure on run-jobs) have stopped working

### DIFF
--- a/qa/tools/apex_algorithm_qa_tools/github_issue_handler.py
+++ b/qa/tools/apex_algorithm_qa_tools/github_issue_handler.py
@@ -21,6 +21,12 @@ from apex_algorithm_qa_tools.scenarios import (
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


jdries (approved):
nice!

HansVRP:
@JanssenBrm I will wait for your approval; since I also believe the apex service account should have a better distinction of what is unstable. 

only in cases where the openEO job fails I would consider it unstable, 

(meta)data changes are often a positive change as we improve the accuracy of openEO

JanssenBrm (approved):

JanssenBrm:
Thanks for the nice feature @HansVRP . Indeed the catalogue has also been updated to be less strict and including a more informative message. Additionally the dashboard now also shows the phase where the benchmark failed. Planned release is tomorrow (15/04) so perfect timing.

HansVRP:
image

seems to work well

will check if I can add some color scheme

PR #469, opened 2026-04-16
Make subswath optional

clausmichele:
@EmileSonneveld currently the burst id was optional but not the subswath. I found out about this trying to call the coherence process with spatial_extent only.


clausmichele:
It's not enough to modify this definition, since if we provide the spatial extent only the burst variable is set and not the subswath https://github.com/cloudinsar/s1-workflows/blob/main/sar/sar_coherence.py

clausmichele:
Once that the PR for the CWL is merged, this can be merged too: https://github.com/cloudinsar/s1-workflows/pull/139

HansVRP:
@EmileSonneveld

EmileSonneveld:
Re-made with fix here: https://github.com/ESA-APEx/apex_algorithms/pull/513

PR #472, opened 2026-04-20
Add sentinel1_sar_slc_preprocessing. 

EmileSonneveld:
https://github.com/eu-cdse/openeo-cdse-infra/issues/838

PR #479, opened 2026-04-24
Sentinel1 sar slc preprocessing v2

EmileSonneveld:
Re make of https://github.com/ESA-APEx/apex_algorithms/pull/472
To test preview, and to be able to merge without potentially breaking links to the MR preview URL


EmileSonneveld:
MR only touches insar related code, and mostly is just adding a new process that was checked in the previous MR. Merging this so we can use 'main' urls.

PR #488, opened 2026-04-29
Cropsar1d first version

Pratichhya:
It is based on the service: https://portal.terrascope.be/catalogue/app-details/10


Pratichhya:
@JanssenBrm could you please suggest if I should update the webapp link in records to a different one as this is a hidden custom process so currently when testing it throws this error: "Can't create function for process 'cropsar1d'. Already exists in Builder class."

Pratichhya:
The execute serive is not working as expected in https://esa-apex.github.io/apex-algorithms-catalogue-web/pr-preview/pr-488/apps/cropsar1d/ does the custom processes with default job options need to be handled differently?

@JanssenBrm @HansVRP could you please suggest me if you know what might be the cause?

algorithm_catalog/vito/cropsar1d/records/cropsar1d.json
```diff
@@ -0,0 +1,124 @@
...
+        },
+        "languages": [
+            {
+                "code": "en-US",
+                "name": "English (United States)"
+            }
+        ],
+        "contacts": [
+            {
+                "name": "Pratichhya Sharma",
```
HansVRP: or Hazan?

algorithm_catalog/vito/cropsar1d/records/cropsar1d.json
```diff
@@ -0,0 +1,124 @@
...
+        ],
+        "license": "proprietary"
+    },
+    "linkTemplates": [],
+    "links": [
+        {
+            "rel": "application",
+            "type": "application/vnd.openeo+json;type=process",
+            "title": "openEO Process Definition",
+            "href": "https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/cropsar1d/algorithm_catalog/vito/cropsar1d/openeo_udp/cropsar1d.json"
```
HansVRP: I do not think this is secure; from within the pg you can fetch the dependencies and potentially get access to the cropsar code

algorithm_catalog/vito/cropsar1d/records/cropsar1d.json
Pratichhya: Sure that sounds even better. I will update it and let her know

algorithm_catalog/vito/cropsar1d/records/cropsar1d.json
Pratichhya: Sorry, not sure if I understand well. Do you mean not exposing the dependency path as in the current case? 

But then how do we pass the dependency as with the current implementation or if there is a different way to access it. 

As for now I cannot define it within the default_job_option for UDP(custom process UDP). https://github.com/Open-EO/openeo-python-driver/issues/483 
or else it will fail to access the cropsar module: j-260513075043453a896c1078b6d75e28

HansVRP: the dependencies should be stored on a path only accesible by openEO (or not shared)

Pratichhya:
@HansVRP @JanssenBrm I think the PR is ready to be merged

HansVRP:
still checking the dependency hosting

HansVRP:
@Pratichhya this one should then be unblocked right?

Pratichhya:
Waiting for package installation on the backend.

Pratichhya:
@HansVRP we are finally ready for this one

algorithm_catalog/vito/cropsar1d/benchmark_scenarios/cropsar1d.json
```diff
@@ -0,0 +1,70 @@
+[
+    {
+        "id": "cropsar1d",
+        "type": "openeo",
+        "description": "CropSAR 1D is a biophysical parameter retrieval algorithm that estimates the Fraction of Absorbed Photosynthetically Active Radiation (FAPAR) for agricultural crops using Sentinel-1 and Sentinel-2 satellite data.",
+        "backend": "openeo.vito.be",
```
HansVRP, line 6: do we need to start using openeo.terrascope.be

algorithm_catalog/vito/cropsar1d/benchmark_scenarios/cropsar1d.json
```diff
@@ -0,0 +1,70 @@
...
+                    "format": "JSON",
+                    "options": {}
+                },
+                "process_id": "save_result",
+                "result": true
+            }
+        },
+        "job_options": {
+            "udf-dependency-archives": [
+                "https://artifactory.vgt.vito.be/artifactory/auxdata-public/marketplace/cropsar1d_dependencies.zip#cropsar_onnx"
```
HansVRP: we probably already want to move this to s2 given that artifactory will be deprecated

algorithm_catalog/vito/cropsar1d/openeo_udp/cropsar1d.json
```diff
@@ -0,0 +1,163 @@
...
+  "default_job_options": {
+    "udf-dependency-archives": [
+      "https://artifactory.vgt.vito.be/artifactory/auxdata-public/marketplace/cropsar1d_dependencies.zip#cropsar_onnxx"
```
HansVRP: see above

algorithm_catalog/vito/cropsar1d/benchmark_scenarios/cropsar1d.json
Pratichhya: I think we will end up with the same error as:
https://github.com/ESA-APEx/apex_algorithms/pull/544#discussion_r3425850767

algorithm_catalog/vito/cropsar1d/openeo_udp/cropsar1d.json
Pratichhya: updated

algorithm_catalog/vito/cropsar1d/benchmark_scenarios/cropsar1d.json
Pratichhya: updated

algorithm_catalog/vito/cropsar1d/openeo_udp/cropsar1d.json
```diff
@@ -0,0 +1,163 @@
...
+      "name": "enddate",
+      "description": "End of the temporal interval and time. The date format is YYYY-MM-DD. If no time is specified, it defaults to 00:00:00. ",
+      "schema": {
+        "type": "string",
+        "subtype": "date",
+        "format": "date"
+      }
+    },
+    {
+      "name": "minimal_input_checks",
```
HansVRP, line 141: wondering if we want to have this one as a user facing parameter

Pratichhya: Yes, it is also a user facing parameter but was implemented for a specific user group based on their request(Pepsico). 

The output can be heavily so is set to false by deafult

PR #490, opened 2026-04-30
proposal for regression testing

HansVRP:
Idea for starting to include regression benchmarks. 

@JanssenBrm I would also need info on how to best expose it such that we can keep a log on the service catalogue


HansVRP:
@JanssenBrm @VictorVerhaert ready to check. I have opted for a more adaptive benchmark where we look at the average and the std. Depending on the nr of successful runs the benchmark becomes more determinantal

HansVRP:
@JanssenBrm @JeroenVerstraelen @VictorVerhaert all feedback is welcome

qa/tools/apex_algorithm_qa_tools/benchmarks.py
```diff
@@ -52,6 +74,144 @@ def collect_metrics_from_results_metadata(
...
+        # Use median + MAD (robust to single spikes/outliers).
+        # MAD scaled by 1.4826 to be consistent with σ for normal data,
+        # so the k schedule (4.0→2.0) stays interpretable as "number of σ".
+        from statistics import median as _median
+
+        median = _median(values)
+        scaled_mad = 1.4826 * _median([abs(v - median) for v in values])
+
+        k = _adaptive_k(min(n, 10))
+        threshold = median + k * scaled_mad
```
VictorVerhaert: Consider adding an absolute buffer for the cost metric here. I wouldn't let a benchmark fail if it suddenly costs 9 instead of 8.

qa/tools/apex_algorithm_qa_tools/benchmark_trends.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+
+    return pyarrow.fs.S3FileSystem(
+        access_key=access_key,
+        secret_key=secret_key,
+        endpoint_override=endpoint,
+    )
+
+
+def load_scenario_history(
```
VictorVerhaert: Consider adding a date cutoff field here which should equal the updated field in the record. That way when a benchmark gets updated it resets the performance tests history.
Not needed and could overcomplicate it, but otherwise we might get a lot of false fails.

VictorVerhaert (commented):
Two small optional comment aimed at trying to prevent false fails. One more question: could you try and run it using github actions and see how it behaves in practice?
Otherwise the pr looks clean

qa/tools/apex_algorithm_qa_tools/benchmarks.py
HansVRP: good point, we do test parameters in various units. so perhaps a standard percentage of the median?

qa/tools/apex_algorithm_qa_tools/benchmark_trends.py
HansVRP: this is already somewhat implemented; it only looks at the 10 last runs. 

I would not add another time limiting parameter

HansVRP:
seems to be working well: 

https://jenkins.vgt.vito.be/job/openEO/job/openeo-apex-benchmarks-handpicked-run/61/

HansVRP:
ended up making some changes seeing the behavior on jenkins. 

So I did add explicit time limitations on data to use and I made cost the sole gating value which can determine a failure. All other values were to volatile, especially for these small benchmarks which hover around 4 credits.

HansVRP:
(blocked by https://github.com/ESA-APEx/apex_algorithms/pull/547)

HansVRP:
@JanssenBrm @soxofaan Could the two of you also take a look at this proposal of regression testing?

In general I look into the merged parquet for the last X months and per usage metric/cost get all valid metrics for succesful runs. 

Based on that I calculate the median and the MAD and convert it into a standard deviation. 

For now the test will fail if the measured cost > median +3.5 x MAD which seemed sensible for some tests I have ran. 

The other metrics fluctuate quite a lot, so for those I only log warnings in case of a failure

soxofaan:
short on time here, so I could only give this a quick look
I'm a bit worried about the design here: if I understand correctly, you are adding hundreds of lines of logic in the main `test_run_benchmark` code path, involving (S3) io, pandas manipulations, which will make the main code path (already known to be brittle and trigger happy), even more brittle and hard to debug.
Can't the performance regression be added as a separate post-processing tool run, working on test suite output?

HansVRP:
@soxofaan most of the code indeed relates to getting a good statistic out of this merger parquet file which can be used to determine there occurred regression or not. Ideally it would be cached somewhere locally such that we would not need to calculate the baseline on runtime each time. 

am I correct that this is what you are proposing as well?

HansVRP:
what I can do is move the baseline calculation to a separate github workflow and trigger it weekly to recompute the baseline. 

in the actual test phase we would read in said file from S3 and check for regression?

soxofaan:
no I'm not talking about caching (I'm not sure there is even something useful to do with caching in a `test_run_benchmark` context),
What I mean is to keep the `pytest ... test_run_benchmark` run/workflow to the bare essentials (for stability and to-the-point results)
and do all the other analysis, trend watching, ... in separate tools or workflows (to allow experimentation, iteration, interactivity)

HansVRP:
makes sense; I'll split both workflows up, let the regression trigger weekly and create separate github issues!

HansVRP:
@JanssenBrm @soxofaan here is the new version, could you take a look? 

What changed: 

1) we invoke the regression test weekly, independent of the integration tests. For now it will only live on github
2) We look back 3 months in the past at the history of each scenario
3) We calculate the median and the MAD (variantion) which are statistically speaking less dependent on outliers. 
4) we calculate an upper and lower limit of median +- 3.5* MAD, I did invoke some safety incase MAD = 0 to avoid too many false positives. 
4) A github ticket is generated if a regression test fails

From testing it became clear that only cost is a good value to check, the others gave way to many false positives. 


example tickets: 

https://github.com/ESA-APEx/apex_algorithms/issues/589

https://github.com/ESA-APEx/apex_algorithms/issues/590

.github/scripts/report_performance_regressions.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python3
```
soxofaan: I would avoid having a scripts folder under `.github`, this conflicts too much with making your code testable or reusable. Being outside of the regular "src' folders also makes it unclear how this code should be handled in terms of packaging, dependencies, ...

As far as I understand, this could just be somewhere under `qa/tools/apex_algorithm_qa_tools/`?

qa/tools/apex_algorithm_qa_tools/benchmark_history.py
```diff
@@ -0,0 +1,117 @@
...
+        scenario_id="max_ndvi_composite",
+        filesystem=fs,
+    )
+"""
+
+import datetime
+import logging
+
+import pyarrow as pa
+import pyarrow.dataset as ds
```
soxofaan: I understand `import ... as ...` is popular in data science, especially for numpy and pandas, but I noticed there is a pet peeve growing in me against this pattern as it obfuscates things massively, unless you're familiar with the aliases.

qa/tools/apex_algorithm_qa_tools/benchmark_regression.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    median = _median(values)
+    mad_raw = _median([abs(v - median) for v in values])
+    mad_scaled_raw = 1.4826 * mad_raw
+    band = max(min_band, k * mad_scaled_raw)
+    mad_scaled = band / k
+    upper_limit = round(median + band, 4)
+    lower_limit = round(max(0.0, median - band), 4)
+    return {
+        "observations": float(len(values)),
```
soxofaan: just curious, why all these `float` casting? aren't the values already float?

Also: why not keep "observations" as an int?

qa/tools/apex_algorithm_qa_tools/benchmark_regression.py
```diff
@@ -0,0 +1,131 @@
+"""Performance regression helpers for benchmark metrics."""
```
soxofaan: you add here two files
- benchmark_history.py
- benchmark_regression.py

the common prefix makes it look like you actually want to create a subpackage ? E.g. something like `benchmark/history.py` and `benchmark/regression.py`, but that is also not ideal as "benchmark" is way to generic in the context of this repo. Maybe better `metrics/history.py` and `metrics/regression.py`?

I'm also confused about `benchmark_history.py`: it's just an added file, but not use, e.g. from `report_performance_regressions.py`. Is that intended?

CHANGELOG.md
```diff
@@ -0,0 +1,47 @@
+# Changelog
```
soxofaan, line 1: Good idea to add a changelog

but there is a time or version dimension missing here to make it clear at what point in time or in which version changes are introduced

soxofaan (commented):
some quick notes

JanssenBrm (approved):

qa/tools/apex_algorithm_qa_tools/benchmark_history.py
HansVRP: done

.github/scripts/report_performance_regressions.py
HansVRP: done

qa/tools/apex_algorithm_qa_tools/benchmark_regression.py
HansVRP: Probably a leftover from when I was debugging and seeing some strange values; will indeed remove!

PR #497, opened 2026-05-07
support new CDSE dev backend URL in test suite

VincentVerelst:
I was recently informed by Peter (can't tag him here) that `https://openeo.dev.waw3-1.openeo-int.v1.dataspace.copernicus.eu/` is the new go-to URL for CDSE dev. However, the test suite didn't accept this one yet.

### DIFF
--- a/qa/benchmarks/tests/conftest.py
+++ b/qa/benchmarks/tests/conftest.py
@@ -112,10 +112,7 @@ def _get_client_credentials_env_var(url: str) -> str:
         return "OPENEO_AUTH_CLIENT_CREDENTIALS_CDSEFED"
     elif hostname == "openeo-staging.dataspace.copernicus.eu":
         return "OPENEO_AUTH_CLIENT_CREDENTIALS_CDSESTAG"
-    elif hostname in {
-        "openeo.dev.warsaw.openeo.dataspace.copernicus.eu",
-        "openeo.dev.waw3-1.openeo-int.v1.dataspace.copernicus.eu",
-    }:
+    elif re.fullmatch(r"openeo\.dev\.([a-z0-9-]+)\.openeo-int\.v1\.dataspace\.copernicus\.eu", hostname):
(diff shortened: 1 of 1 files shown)


soxofaan:
maybe it's better in this case to write a regex instead of trying to have an exact list ?

VincentVerelst:
> maybe it's better in this case to write a regex instead of trying to have an exact list ?

@soxofaan, how much freedom should this regex then have? Don't you have the risk of allowing a non-existing backend URL when not using these allow lists?

soxofaan:
@pvbouwel  do you have a suggestion for a regex to avoid hardcoding CDSE dev hostnames like 
```
    elif hostname in {
        "openeo.dev.warsaw.openeo.dataspace.copernicus.eu",
        "openeo.dev.waw3-1.openeo-int.v1.dataspace.copernicus.eu",
    }:
```

pvbouwel:
So the only form that should be supported for our cloud environments is

```
service.env.region.<project>[-int].v1.dataspace.copernicus.eu
```

For OpenEO endpoint service is "openeo"
environment is "dev|stag|prod"
region is the region name within the cloud provider e.g.  "waw3-1|waw4-1|eu-nl|eu-west-1|..." 
project is like for what the group of endpoints is setup. So can be OpenEO but if there is a project with its own stack it could be any value (currently always openeo) 

So openeo.dev.warsaw.openeo.dataspace.copernicus.eu should anyway be removed as there should not be a use case that cannot be served by openeo.dev.waw3-1.openeo-int.v1.dataspace.copernicus.eu and that one should be preferred.

soxofaan:
@VincentVerelst 

so ideally the regex should be something like

```python
elif re.fullmatch(r"openeo\.dev\.([a-z0-9-]+)\.openeo-int\.v1\.dataspace\.copernicus\.eu", hostname):
```

JanssenBrm:
@soxofaan  - Can this PR be merged?

soxofaan (approved):

soxofaan:
done

PR #501, opened 2026-05-15
Add FORCE to the algorithm catalog

Hneuschmidt:
This PR introduces [FORCE](https://force-eo.readthedocs.io/en/latest/) as an algorithm to the APEx algorithm catalogue.
The source for the FORCE EO Application package is contained in the [apex-force-openeo](https://github.com/bcdev/apex-force-openeo) repository

> [!IMPORTANT]
> Integration work on FORCE is still ongoing so details may change

It is still early to make FORCE fully available. However, I would like to start the process of integrating the record to make sure we have all metadata ready by the time the integration work is done.

@JanssenBrm David Frantz (author and maintainer of FORCE) and I were wondering who to mention as the "algorithm provider" in this case. Unlike (all?) the other examples in the APEx algorithm catalogue, this software is provided by a researcher at a university with no involvement by a company. We could add e.g. the university, the group or David as a person as the algorithm provider. Do we have a policy for these cases?

A few thoughts to consider:

- As I understand, David owns the IP
- University / Group association may change in the future, the IP moves with the owner


JanssenBrm:
> @JanssenBrm David Frantz (author and maintainer of FORCE) and I were wondering who to mention as the "algorithm provider" in this case. Unlike (all?) the other examples in the APEx algorithm catalogue, this software is provided by a researcher at a university with no involvement by a company. We could add e.g. the university, the group or David as a person as the algorithm provider. Do we have a policy for these cases?

There are currently no policies in place for this. The role field is also a free-text field and you can add multiple contacts. I suggest to add David Frantz with the roles author and maintainer. If David also wants to mention the university or group this is also fine.

Hneuschmidt:
Merge should wait until release 1.0 of the FORCE EOAP. However, the changes can be already reviewed @JanssenBrm .

Hneuschmidt:
@JanssenBrm I believe unit test failures are now caused by pre-existing failures from the main branch.

A few things remain to be done:

- [ ] The schema requires an "application" link, I assume this is what will be opened by the "Execute Service" button. I have now added the example notebooks there, but there is no guarantee yet that they would run on Jupyterlite. For FORCE level 2 we could somewhat reasonably add a UDP that includes a query, but the combination of querying with openEO and FORCE level 2 is much more limited that FORCE level 2 itself, so I am not sure we should do that. What do you think? 
- [ ] The FORCE custom processes `force_level2` and `force_tsa` rely on the latest release of https://github.com/bcdev/apex-force-openeo, we should only merge once 1.0 has been released (I am working on that currently).
- [ ] I have added a description markdown document to both processes as I have seen in other places. I am not sure how this must be added to appear as the rendered description.
- [ ] It would be great to add the parameters of the `force_level2` and `force_tsa` processes. Is it possible to link "built-in" processes like this (instead of a process_graph.json) somehow?

Hneuschmidt:
Dear @JanssenBrm, the FORCE integration is complete and v1.0.0 has now also been released. Could you review and give feedback on the questions above this week?

PR #508, opened 2026-05-21
entry rayview

HansVRP:


HansVRP:
@JanssenBrm PR is ready to be merged; few failing unit test which are unrelated

JanssenBrm:
Thanks for the record @HansVRP! Some minor feedback:

- The description on the record is very minimal. I can see algorithm_catalog/ucl/solar_pv_detection/openeo_udp/solar_pv_detection_description.md but the algorithm catalogue is reading the description from the [UDP](https://raw.githubusercontent.com/ray-climate/solar_openEO/main/openeo_udp/process_graph/solar_pv_detection_udp.json) itself. This ensures that the description is consistent across tooling using the UDPs.
- The `bbox` parameter is defined as an `object`. This is causing strange behaviour in the openEO web editor and will also cause errors in other tools. Can this be updated to use the openEO [bounding box](https://open-eo.github.io/openeo-python-client/api.html#openeo.api.process.Parameter.bounding_box) parameter?
- Similar as above for the end date, maybe it is better to use the [date](https://open-eo.github.io/openeo-python-client/api.html#openeo.api.process.Parameter.date) parameter to ensure that a nice datepicker is shown?

Out of my interest: there is an `end_date` parameter, so I would also expect a start date? Unless I've missed it, the documentation does not clearly state what will happen. Does it process the full archive until the `end_date`?

PR #514, opened 2026-05-27
entry delinieate anything

HansVRP:


JanssenBrm:
Sorry for my late response @HansVRP. Thank you for integrating the service, my comments are the same as with the UCL:

- The parameters are now defined as general objects, making the integration within external tools not so user friendly.
- The description is quite limited, not really explaining what the service does and how it works. 

If those could be taken care of, I believe we can merge the record into the catalogue.

PR #542, opened 2026-06-11
Remove null. Pointed out by McGill.

EmileSonneveld:
For a long term solution, this could be avoided when automatically generating the json based on CWL.


EmileSonneveld:
@copilot review

EmileSonneveld:
`test_service_record_validation` failed in gh actions, just like other branches. Will merge

PR #544, opened 2026-06-12
Cropsar2d 

Pratichhya:
for the service hosted in: https://portal.terrascope.be/catalogue/app-details/86


algorithm_catalog/vito/cropsar2d/benchmark_scenarios/cropsar2d.json
```diff
@@ -0,0 +1,70 @@
+[
+    {
+        "id": "cropsar2d",
+        "type": "openeo",
+        "description": "CropSAR 1D is a biophysical parameter retrieval algorithm that estimates the Fraction of Absorbed Photosynthetically Active Radiation (FAPAR) for agricultural crops using Sentinel-1 and Sentinel-2 satellite data.",
+        "backend": "openeo.vito.be",
```
HansVRP: does this need to become openeo.terrascope.be and or cdse?

algorithm_catalog/vito/cropsar2d/openeo_udp/cropsar2d.json
```diff
@@ -0,0 +1,152 @@
...
+  "default_job_options": {
+    "udf-dependency-archives": [
+      "https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies/onnx_deps_python311.zip#onnx_deps"
```
HansVRP: I think its recommended to leave out the swift/v1

Pratichhya: updated

algorithm_catalog/vito/cropsar2d/benchmark_scenarios/cropsar2d.json
Pratichhya: I did a test last week with the apex-handler where it didn't support the openeo.terrascope.be backend, so thought to leave it out, but seems to be the same case here: https://github.com/ESA-APEx/apex_algorithms/actions/runs/27668299704/job/81826818361

Should I continue with openeo.vito.be or wait for the fix?
And it should be terrascope indeed not cdse for the given custom processes and dependencies

HansVRP: depends on whether this will break in a few weeks. Can you follow this up with the python team? it is not a blocker however

HansVRP (approved):

PR #547, opened 2026-06-19
feat: added automatic merging of parquet files

JanssenBrm:
This PR includes a script that would automatically merge the benchmarking parquet files after each benchmarking run. This approach ensures that:

- The information on the APEx Algorithm Dashboard is always up-to-date
- The benchmark notifications have an up-to-date historic reference of previous runs


HansVRP (commented):
Just a few questions from my end. I assume this writes away both saved as failed jobs. 

does the current set-up allow for 'partial' appending as well where e.g. the usage logs are unavailable. 

Secondly is it worth adding some 'gating' to which jobs we do write away? do we want entries is start_job failed due to network issues?

JanssenBrm:
Thanks for the quick response Hans!

* Yes, it writes both the saved and failed jobs 
* It just takes the parquet content as-is and appends it, so if the usage logs are unavailable, metrics won't be appearing in the merged version as they are probably missing in the initial parquet. 
* Filtering out these scenarios could make sense to not clutter the results with false negatives. However, I believe this logic should not be part of the merging component, suggesting to keep in the benchmarks themselves.

HansVRP:
sure, no comments from my side then apart from performing a test run on via https://[jenkins.vgt.vito.be](https://jenkins.vgt.vito.be/job/openEO/job/openeo-apex-benchmarks-handpicked-run/)/job/openEO/job/openeo-apex-benchmarks-handpicked-run/ or a github run and validating that it has the intended behavior. If the test has already been run, then link it in the PR here

JanssenBrm:
Thanks @HansVRP! Added actual unit testing to test the behaviour of the merge_parquet functions. After reconsidering, it might be overkill to always check this based on an actual benchmark result, as a unit test would cover the same behaviour.

qa/benchmarks/merge_parquet.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Read the partitioned metadata
+    _log.info(f"Reading parquet files from S3 bucket: {s3_bucket}")
+    read_partitioning = pyarrow.dataset.partitioning(
+        schema=pyarrow.schema(fields=[("test:start:YYYYMM", pyarrow.string())]),
+        flavor=None,
+    )
+
+    table = pyarrow.parquet.read_table(
+        f"{s3_bucket}/metrics/v1/metrics.parquet",
```
soxofaan: maybe expose the key used here instead of hardcoding, e.g. the pytest run has:

https://github.com/ESA-APEx/apex_algorithms/blob/9a13aaf633de8330b4b6058bfac577464efca777/.github/workflows/benchmarks.yaml#L61

qa/benchmarks/merge_parquet.py
```diff
@@ -0,0 +1,126 @@
...
+    # Initialize S3 filesystem
+    s3fs = pyarrow.fs.S3FileSystem(
+        access_key=s3_client, secret_key=s3_secret, endpoint_override=s3_endpoint, region=s3_region
+    )
+
+    # Read the partitioned metadata
+    _log.info(f"Reading parquet files from S3 bucket: {s3_bucket}")
+    read_partitioning = pyarrow.dataset.partitioning(
+        schema=pyarrow.schema(fields=[("test:start:YYYYMM", pyarrow.string())]),
+        flavor=None,
```
soxofaan: I would add comment about what "flavor=None" means, because it's a bit obscure what this is about. e.g. from my original notebook:
```
# Partitioning flavor `None`` means DirectoryPartitioning (see https://github.com/apache/arrow/issues/43863)
```

qa/unittests/tests/pytest/test_merge_parquet.py
```diff
@@ -0,0 +1,235 @@
...
+_spec = importlib.util.spec_from_file_location(
+    "merge_parquet",
+    Path(__file__).parents[3] / "benchmarks" / "merge_parquet.py",
+)
+_mod = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_mod)
+merge_parquet_files = _mod.merge_parquet_files
+
+# S3 path where the benchmark suite writes its metrics.
+_S3_METRICS_KEY = "metrics/v1/metrics.parquet"
```
soxofaan: maybe use something intentionally different from what is used in production to avoid accidental messing with prod data?

qa/unittests/tests/pytest/test_merge_parquet.py
```diff
@@ -0,0 +1,235 @@
...
+        assert df.loc["biopar"]["test:outcome"] == "failed"
+        assert df.loc["biopar"]["job_id"] == "j-def456"
+
+    def test_s3_output_row_count_and_content(
+        self, s3fs, s3_bucket, moto_server, tmp_path
+    ):
+        """Merging to an S3 output path produces a readable dataset with all source rows."""
+        _write_test_data_to_s3(s3fs=s3fs, s3_bucket=s3_bucket, rows=[_ROW_PASSED, _ROW_FAILED])
+
+        output_s3_path = f"{s3_bucket}/metrics/v1/metrics-merged.parquet"
```
soxofaan: same: possible to avoid using same paths as used in prod?

soxofaan (commented):
looks already fine as-is I think (it's been a while I played with pyarrow in practice), so just some superficial notes

PR #567, opened 2026-07-09
world_agrocommodities not private anymore

VincentVerelst:
Now that the workflow is mature, we'd like to onboard WAC to the APEx algorithm catalogue as an on-demand service. 

@JanssenBrm , are these changes sufficient to do so?


JanssenBrm:
Thanks @VincentVerelst  - This is indeed sufficient. However, looking at the record itself (https://esa-apex.github.io/apex-algorithms-catalogue-web/pr-preview/pr-567/apps/world_agrocommodities/), I noticed the following:

* The description is rather limited. We are reading the description from the UDP `description` field. It might benefit from a bit more information.
* There is no link to the `webapp` link in the record. This means that the user has no direct way of using the service directly from the catalogue. Do you think you might be able to (at a minimum) add a webapp link (example: https://github.com/ESA-APEx/apex_algorithms/blob/main/algorithm_catalog/vito/biopar/records/biopar.json#L140)? Adding a link to a notebook would be more increase the experience.

VincentVerelst:
@JanssenBrm , added a webapp record and updated the description!

HansVRP (approved):

JanssenBrm (approved):

PR #571, opened 2026-07-09
Adds LAI UDP

firzaariany:
Adds an algorithm that computes Leaf Area Index (LAI) from Sentinel-2 L2A imagery as an openEO UDP. The openEO process is adapted from a SentinelHub LAI evalscript, porting an SNAP neural network (8 spectral bands + 4 angular inputs) to estimate the one-sided green leaf area per unit of land area (m²/m²). The algorithm produces raw LAI index values as output.

The openEO process is running on Copernicus Data Space over the spatial extent of Cropland Plain, Istria Coast, Croatia on 2025-05-10 to 2025-05-12.


firzaariany:
@HansVRP, similar to PR #572, the unit tests in this PR are also failing because of temporary 504 gateway timeouts rather than from records misalignment with requirements. I couldn't find the re-run button so could you help me how to solve this? Perhaps you could re-run, and I think it could solve the network issue.

HansVRP (approved):

PR #572, opened 2026-07-09
Add UDP for visualizing Plant Senescence Reflectance Index (PSRI)

firzaariany:
Adds an algorithm that computes the PSRI from Sentinel-2 L2A imagery as an openEO UDP. The openEO process is adapted from a SentinelHub PSRI evalscript, to track the progression of leaf senescence and natural vegetation aging, then rendering the result to [0, 1].

The openEO process is running on Copernicus Data Space over the spatial extent of Čepić plain, Istria, Croatia on 2025-05-10 to 2025-05-12.


firzaariany:
Hi @HansVRP, the unit tests workflow failed with transient GitHub API 502 errors (Bad Gateway) when hitting the Contents API, causing two tests to fail: `TestGithubAlgorithmRepository::test_list_algorithms` and `::test_get_algorithm`.

Could you please re-run the failed job? This looks like a temporary upstream error and a re-run will likely pass. Thanks!

HansVRP (approved):

PR #580, opened 2026-07-10
Fixes to the devseed UDPs 

JanssenBrm:
Currently some of the UDPs did not use the bounding box and time parameters, resulting in always the same image, independent of the parameters that the user specified.


JanssenBrm:
@firzaariany  - Could have you have a quick look to see if my changes do not break the current flows?

firzaariany:
@JanssenBrm I'm on it

PR #593, opened 2026-07-10
Fix: UDPs not working

firzaariany:
Related to #580 

I needed to fix some UDPs to test in the browser. They all work well and standardize now. I confirmed that they work as intended and not bounded to a specific/hardcoded spatial and temporal bounds. 

Even though I'm quite sure that re-target the branch to `ESA-APEx:main` would include all the changes from `ESA-APEx:fixes_devseed_udps` (2 commits) and `developmentseed:fixes_devseed_udps` (17 commits), I'm still hesitant to do it. I set `developmentseed:fixes_devseed_udps` to track the upstream branch with the same name. Need input on this

Issues:
- [x] Wildfire Visualization: Removes `mask_polygon` process with hard-coded bounds
- [x] Atmospheric Resistant Vegetation Index: Resolves output format definition
- [x] Enhanced Vegetation Index: Removes `mask_polygon` process with hard-coded bounds
- [x] Kernel NDVI: Add file format and cloud filter
- [x] LAI: Removes `mask_polygon` process with hard-coded bounds
- [x] PSRI: Resolves empty output file format and cloud filter

### DIFF
--- a/algorithm_catalog/developmentseed/atmospheric_resistant_vegetation_index/openeo_udp/atmospheric_resistant_vegetation_index.json
+++ b/algorithm_catalog/developmentseed/atmospheric_resistant_vegetation_index/openeo_udp/atmospheric_resistant_vegetation_index.json
@@ -9,6 +9,22 @@
           "B8A"
         ],
         "id": "SENTINEL2_L2A",
+        "properties": {
+          "eo:cloud_cover": {
+            "process_graph": {
+              "lte1": {
+                "process_id": "lte",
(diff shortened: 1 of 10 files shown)


firzaariany:
@JanssenBrm hi, sorry I had to fix some UDPs in a forked branch as they were not working cleanly. The problem wasn't due to removing the usage of the default parameters as you did in #580, but rather due to some processes were not integrated in the UDPs (e.g. `save_result`). What would be a cleaner way to merge this?

PR #596, opened 2026-07-13
Ard terravision

### LINKED ISSUE #409: Integration of TerraVision Spectral Indices
Within the TerraVision project, a consortium partner is creating spectral indices using openEO that are relevant for the mining industry. These are also interesting to incorporate as services on the APEx Algorithm Catalogue.

More information is to be provided over time by the TerraVision project. They are currently experimenting with the different indices and their implementation in openEO.

Pratichhya:
First version of including two UDPs implemented as a part of the Terravision project: https://terravision-project.eu/
These include simple indices calculations that are used within the project, namely:
* BSI
* AMWI

related PR: https://github.com/ITA-TECNOLOGIA/Ard-Terravision/pull/2


Pratichhya:
I think now it is failing mostly due to a timeout. @JanssenBrm in the meantime, please let me know if you have any suggestion or feedback

JanssenBrm:
Hi @Pratichhya,

This looks very nice, I just did a small fix to the formatting of the amwi record as it was causing issues. Added a test to prevent this in the future.

My only comment on is that the UDP description is quite "basic". It is not really explaining what the index is about and how to interpret the values in the result. It would be great if these could be added as this will improve the service record quite a bit.

Pratichhya:
@JanssenBrm thank you for the check. I updated the description in benchmark json files to reflect in the main page

JanssenBrm:
Thank you for the quick response @Pratichhya  but I'm afraid the update in the benchmark scenario is not sufficient. The APEx Algorithm Catalogue reads the description from the actual UDP: https://raw.githubusercontent.com/ITA-TECNOLOGIA/Ard-Terravision/b848749026d3397cd89969e549712e21e992d4a7/src/main/python/utils/openeo_udp/amwi.json. So the description should be updated in that JSON (and probably the script generating the JSON).

Pratichhya:
Thank you for the pointer Bram and sorry for the trouble.

I updated the description of the UDP itself as well. However, it is not yet reflected in https://esa-apex.github.io/apex-algorithms-catalogue-web/pr-preview/pr-596/apps/amwi/#execution-information 

not sure if I am missing anything else.

PR #609, opened 2026-07-15
Remove the save_result from the main UDP

### LINKED ISSUE #608: Remove save_result from OBSGESSION LAI UDP
See https://github.com/ESA-APEx/projectmanagement/issues/166

JanssenBrm:
In the current implementation, the `save_result` step is included in the full UDP definition. Whenever a user uses the `execute_batch` function without changes, it always downloads 2 files. This is causing the following confusion:

* Users need to explicitly add an `auto_add_save_result=False`, which requires them to either look at the UDP or only realise (if they do) after executing the service. It is not documented, except for the notebook example.
* This step forces users to use the GTIFF format. If users do select a different format in the `execute_batch` the GTIFF is downloaded anyway.
* I'm not sure how this will behave when we are chaining this UDP to another one.

Nevertheless, I do see the usefulness of adding the metadata to the output files and maybe there is a workaround for this.


JanssenBrm:
@HansVRP / @Pratichhya  - Feel free to share your thoughts on how we can best approach this. Right now it is quite unclear that the user needs to add this `auto_add_save_result=False` to their process. 

The issue was raised by the BiaB integration where they simply read out a UDP definition from the APEx Algorithm Catalogue in an automated way, so there is no real way for them figuring out whether to add this parameter to the `execute_batch` yes or no (same applies for all other users btw).

mbuchhorn:
@JanssenBrm @Pratichhya I would be fine with removing the save step if you can keep the metadata somehow in the final product the user downloads (the metadata is compartible with Gtif and NetCDF.... zar not tested)

Pratichhya:
From the user side, I think @JanssenBrm already implemented an alternative approach for handling metadata in the notebook workflow here:

https://github.com/ESA-APEx/apex_algorithms/blob/608-remove-save_result-from-obsgession-lai-udp/algorithm_catalog/obsgession/udp_obsgession_w23_lai/notebooks/execute-obsgession-lai-udp.ipynb

To further investigate how to update the output metadata outside the save_result node while still ensuring it is preserved in the final downloaded product (e.g., GTiff and NetCDF outputs), it might be worth creating a separate issue to explore metadata handling: https://github.com/ESA-APEx/apex_algorithms/issues/611

Pratichhya (approved):

